PR #139, opened 2015-03-26
Update mozjs_sys to its new name.

Ms2ger:


metajack:
Are we mass renaming all the -sys packages now? Why is this needed?

jdm:
Support for hyphens in crate names is going away.

PR #144, opened 2015-04-24
Upgrade to SM 39

michaelwu:
Some things have been implemented in rust.rs. Where possible, things are being implemented using the structs generated by my bindgen fork (which I will call bindgen++ from now on).

jsapi.rs has been hand edited a bit. Note that jsapi.rs is generated from jsfriendapi.hpp so we can eliminate jsfriendapi.rs entirely. Making jsfriendapi.rs work with with jsapi.rs was a bit difficult, so I combined them to sidestep the problem.

Here are the things I had to hand edit in jsapi.rs:
- Adding "unsafe impl Sync for " for structs that are statically defined.
- Adding "pub enum Foo {}" for types not defined in js headers. (usually mfbt)
- Adding "#[derive(Copy)]" where necessary to those defines.
- Removing derive copy from autogenerated structs that don't support it because bindgen++ doesn't fully define all the structs it depends on.
- Trimming away generics for structs that didn't get converted successfully. This can happen for a number of reasons:
  - bindgen++ can only convert templates that work on types
  - bindgen++ can't convert the classes that a struct/class might inherit from
- 64 bit issues:
  - Certain types need to be converted back to size_t since bindgen++ currently replaces them with u32 or u64. This will be fixed.
  - Some mangled function names are different between 32 bit and 64 bit, usually because of 64 bit types and size_t/long. In these cases, functions are hand edited to add 64bit versions. Might be fixed in the future, but for now we can only detect these problems when code doesn't link.
- Spidermonkey checks for the address of JS_GlobalObjectTraceHook in globals. For now, I made the mangled name public so we can grab the real function address, but we can probably do a little bit better.
- References to JSVal are replaced with Value, since jsapi.rs and jsval.rs both define JSVal. In jsapi.rs, JSVal is just a typedef to Value, so we can drop JSVal in jsapi.rs.
- bindgen++ doesn't support preprocessor defines at the moment, so some defines were manually added.
- JSJitMethodCallArgs was manually defined.
- JS_GetRuntime wasn't converted by bindgen++, so it was manually added. Need to investigate.
- Structs with unions generated by bindgen++ are generally unusable for statically defined things, so the necessary structs were modified to be more friendly for that case.


jdm:
That sounds exactly like what I ran into for Root<T> vs Root<*mut T> in the previous upgrade attempt.

PR #146, opened 2015-05-09
Test on Travis.

Ms2ger:


jdm:
I'm fine with this, but do we want to get the tests passing first?

PR #151, opened 2015-05-28
Get dependency directory from cargo metadata

asbaker:
Part of servo/servo#6088. Depends on servo/mozjs#38  @SimonSapin

### DIFF
--- a/makefile.cargo
+++ b/makefile.cargo
@@ -20,8 +20,7 @@ ifneq (,$(CARGO_FEATURE_DEBUGMOZJS))
 	CFLAGS += -g -O0 -DDEBUG -D_DEBUG
 endif
 
-MOZJS_OUTDIR = $(shell find $(OUT_DIR)/../.. -name 'mozjs_sys-*' -type d)/out
-CFLAGS += -I$(MOZJS_OUTDIR)/dist/include
+CFLAGS += -I$(DEP_MOZJS_OUTDIR)/dist/include
 
 .PHONY: all
 all: $(OUT_DIR)/libjsglue.a


makefile.cargo
```diff
@@ -20,8 +20,7 @@ ifneq (,$(CARGO_FEATURE_DEBUGMOZJS))
 	CFLAGS += -g -O0 -DDEBUG -D_DEBUG
 endif
 
-MOZJS_OUTDIR = $(shell find $(OUT_DIR)/../.. -name 'mozjs_sys-*' -type d)/out
-CFLAGS += -I$(MOZJS_OUTDIR)/dist/include
+CFLAGS += -I$(DEP_MOZJS_OUTDIR)/dist/include
```
SimonSapin: I think this is missing the `out` component.

asbaker: @SimonSapin I checked it out and it looks like the /out component is part of DEP_MOZJS_OUTDIR from mozjs: ```cargo:outdir=/Users/abaker/Source/rust-mozjs/target/debug/build/mozjs_sys-476fe81cd282b5dc/out```

Here is the result of ```@echo```'ing the old variable and the new variable from rust-mozjs's makefile.cargo:
```MOZJS_OUTDIR /Users/abaker/Source/rust-mozjs/target/debug/build/js-892e1c7dbc7e3cb5/out/../../mozjs_sys-476fe81cd282b5dc/out```
```DEP_MOZJS_OUTDIR /Users/abaker/Source/rust-mozjs/target/debug/build/mozjs_sys-476fe81cd282b5dc/out```

Looks like CI is just being cranky because the build ran before mozjs was merged to master. It didn't pick up a value for $(DEP_MOZJS_OUTDIR) at all: ```-I/dist/include```

SimonSapin: Sorry, you’re right. `/out` is indeed included in the environment variable. I restarted the Travis build and it’s green now.

PR #159, opened 2015-06-21
Implement finalizeInBackground.

Ms2ger:
This prevents SM from silently finalizing objects off the JSRuntime's thread.


metajack:
@Ms2ger Why is this needed?

Ms2ger:
Without this, we end up calling `Document::drop` off the script thread, which introduces data races.

metajack:
Out of curiosity, does Firefox let the JS engine finalize in the background? How do they deal with it?

Ms2ger:
Gecko doesn't allow it either.

PR #162, opened 2015-06-25
Revert "Fix build in Raspberry Pi2."

Ms2ger:
Reverts servo/rust-mozjs#157


Ms2ger:
It breaks the Servo build on android/gonk

Issue #164, opened 2015-07-01
Errors are not being reported


zofrex:
Out of the box, the registered error reporting function is not being called for me. I've narrowed it down to these two lines in Runtime::new():

``` rust
            (*contextopts).set_dontReportUncaught_(true);
            (*contextopts).set_autoJSAPIOwnsErrorReporting_(true);
```

Unless both of these options are set to false, the error reporting function doesn't get called on error.

Ms2ger:
`autoJSAPIOwnsErrorReporting` should definitely be true for Servo. Paging @bholley.

bholley:
Yes. If you want to avoid an error reporting nightmare, Servo should do the following:
1: Set autoJSAPIOwnsErrorReporting and dontReportUncaught everywhere (hopefully on the single JSContext that Servo presumably has).
2: Register a "WarningOnlyErrorReporter", that behaves similarly to the one set by AutoJSAPI::TakeOwnershipOfErrorReporting.
3: Implementing an RAII class like AutoJSAPI, which handles errors when it gets popped off the stack.

KishanV:
How can I enable error reporting on current rust-mozjs version....

Ms2ger:
SpiderMonkey no longer calls a callback to report errors, just warnings (`SetWarningReporter`). You'll need to handle errors yourself when appropriate; there's [some code in Servo](https://github.com/servo/servo/blob/6c3e94805f74e78b091b37fe374af12ad2249528/components/script/dom/bindings/error.rs#L200) you might be able to reuse.

Issue #134, opened 2015-02-18
Create a sample applications


llacroix:
As there is not yet any kind of documentation on how to use the rust-mozjs binding of the jsapi. It would be cool to have something that mimmick the sample jsapi helloworld app from mozilla. (By documentation I mean user guide and not just reference documentation).

As some things are a bit different, I wasn't able to build an application that works like the minimal application here:

https://developer.mozilla.org/en-US/docs/Mozilla/Projects/SpiderMonkey/JSAPI_User_Guide

jdm:
I agree that an example app would be nice to have, but I have no time to create it myself. I'm willing to help someone else set it up, however!

jdm:
And yes, if that wiki page is up to date, it probably uses APIs that are 2.5 years newer than the ones available in this repository.

llacroix:
I believe the helloworld program there has been the same for at least 2 years.

zofrex:
Proposing closing this now that #165 has landed

llacroix:
Cool thanks

PR #165, opened 2015-07-01
Add a basic example

zofrex:
I am sure this still needs work, but it's a starting point for a basic example - #134. Credit to jdm, Ms2ger, and mbrubeck for helping me get this far!

Things that are missing compared to the SpiderMonkey documentation:

* JS_InitStandardClasses
* Error handling (see #164)

### DIFF
--- a/examples/callback.rs
+++ b/examples/callback.rs
@@ -1,0 +1,76 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+extern crate js;
+extern crate libc;
+
+use std::ffi::CStr;
+use std::ptr;
+use std::str;
+
+use js::{JSCLASS_RESERVED_SLOTS_MASK,JSCLASS_RESERVED_SLOTS_SHIFT,JSCLASS_GLOBAL_SLOT_COUNT,JSCLASS_IS_GLOBAL};
+use js::jsapi::_Z24JS_GlobalObjectTraceHookP8JSTracerP8JSObject;
+use js::jsapi::{CallArgs,CompartmentOptions,OnNewGlobalHookOption,Rooted,Value};
+use js::jsapi::{JS_DefineFunction,JS_Init,JS_NewGlobalObject,JS_EncodeStringToUTF8,JS_ReportError};
+use js::jsapi::{JSAutoCompartment,JSAutoRequest,JSContext,JSClass};
+use js::jsval::UndefinedValue;
+use js::rust::Runtime;
+
+static CLASS: &'static JSClass = &JSClass {
+    name: b"test\0" as *const u8 as *const libc::c_char,
+    flags: JSCLASS_IS_GLOBAL | ((JSCLASS_GLOBAL_SLOT_COUNT & JSCLASS_RESERVED_SLOTS_MASK) << JSCLASS_RESERVED_SLOTS_SHIFT),
+    addProperty: None,
+    delProperty: None,
+    getProperty: None,
+    setProperty: None,
+    enumerate: None,
+    resolve: None,
+    convert: None,
+    finalize: None,
+    call: None,
+    hasInstance: None,
+    construct: None,
+    trace: Some(_Z24JS_GlobalObjectTraceHookP8JSTracerP8JSObject),
+    reserved: [0 as *mut _; 25]
+};
+
+fn main() {
+    unsafe {
+        JS_Init();
+    }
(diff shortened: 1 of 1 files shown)


zofrex:
This example works and doesn't crash, but with SM assertions on, it fails:

```$ cargo run --features debugmozjs --example callback
Assertion failure: cx->runtime()->requestDepth || cx->runtime()->isHeapBusy(), at /Users/zofrex/.cargo/git/checkouts/mozjs-06d7f04b6dbb8a8e/master/mozjs/js/src/jscntxt.cpp:1144
An unknown error occurred```

**Comment by michaelwu** [comment]
Add in a `JSAutoRequest`

**Comment by zofrex** [comment]
Thanks! Getting a different one now, can't figure this one out...

```
Assertion failure: clasp->trace == JS_GlobalObjectTraceHook, at /Users/zofrex/.cargo/git/checkouts/mozjs-06d7f04b6dbb8a8e/master/mozjs/js/src/vm/GlobalObject.cpp:243
```

Going to take a break and revisit this tomorrow.

**Comment by michaelwu** [comment]
It wants JS_GlobalObjectTraceHook in trace, but you have trace set to None. Note that you'll need to use the mangled version of that.

**Comment by Ms2ger** [comment]
Don't forget the MPL2 header.

**Comment by zofrex** [comment]
Thanks! Bit horrible that you can't use the JS_GlobalObjectTraceHook wrapper :(

Almost there, it's now failing an assert at the end of evaluate_script in rust.rs:

`Assertion failure: mStatementDone, at ../../dist/include/mozilla/GuardObjects.h:100`

Looks like it's complaining about a guard object used as temporary, but I can't find one...

**Comment by Ms2ger** [comment]
`Rooted::new(context, global).handle()` looks highly suspicious

**Comment by jdm** [comment]
https://github.com/servo/rust-mozjs/pull/165/files#diff-490a245ae6cc96c180d8cb9a9a8eaaa6R50 looks wrong to me - the RootedObject will only be a temporary that will live as long as that statement, and the Handle will point to invalid memory.

**Comment by michaelwu** [comment]
You're making a handle in the puts functions. In general, you can't make manually make handles unless you know exactly what you're doing.

I don't think I can help with the guard object issue without more information to figure out what object the guard object is embedded in.

**Comment by zofrex** [comment]
I believe I've fixed both the rooting / handle issues. Michael - I don't know how to figure out which object the guard object is embedded in. I tried adding a breakpoint on guard object creation but that was far too noisy to be useful.

**Comment by michaelwu** [comment]
A stack on that assertion would help.

**Comment by zofrex** [comment]
https://gist.github.com/zofrex/a7832f69c06b08e0135e

**Comment by michaelwu** [comment]
This might be an issue with the C++ part of the glue being built without debug/debugmozjs enabled. Try clobbering target/ and building from scratch.

**Comment by zofrex** [comment]
That works! (And also explains why this code worked without crashing on my other machine...)

Alright, the code compiles and runs with no errors, even with debugmozjs enabled. What's next?

File: examples/callback.rs [hunk 1]
```diff
@@ -0,0 +1,76 @@
...
+
+    unsafe {
+        let h_option = OnNewGlobalHookOption::FireOnNewGlobalHook;
+        let c_option = CompartmentOptions::default();
+        let _ar = JSAutoRequest::new(context);
+        let global = JS_NewGlobalObject(context, CLASS, ptr::null_mut(), h_option, &c_option);
+        let global_root = Rooted::new(context, global);
+        let global = global_root.handle();
+        let _ac = JSAutoCompartment::new(context, global.get());
+        JS_DefineFunction(context, global, CString::new("puts").unwrap().as_ptr(), Some(puts), 1, 0);
```
jdm: Instead of CString, this could probably be `b"puts\0".as_ptr()`

examples/callback.rs
```diff
@@ -0,0 +1,76 @@
...
+        let javascript = "puts('Test Iñtërnâtiônàlizætiøn ┬─┬ノ( º _ ºノ) ');".to_string();
+        let _ = runtime.evaluate_script(global, javascript, "test.js".to_string(), 0);
+    }
+}
+
+unsafe extern "C" fn puts(context: *mut JSContext, argc: u32, vp: *mut Value) -> u8 {
+    let args = CallArgs::from_vp(vp, argc);
+
+    if args.argc_ != 1 {
+        JS_ReportError(context, CString::new("puts() requires exactly 1 argument").unwrap().as_ptr());
```
jdm: Same as above.

zofrex:
It needs a `as *const i8` or `as *const libc::c_char` if I do it the `b"something.as_ptr()` way. I have no opinion on which is less ugly.

jdm:
I'd go with c_char.

PR #167, opened 2015-07-02
Allow to override CC and AR for cross compilation.

mmatyas:
Requires servo/servo#6531
Part of servo/servo#6327

[Review on Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/167)


mmatyas:
This is the same as the previously reverted Raspberry fix, but now the gonk side is also fixed.

mmatyas:
Updated, works for both ARM and Android.

Issue #161, opened 2015-06-22
cargo test fails to link


Ms2ger:
... and we gate on it now.

Ms2ger:
CC @metajack @larsbergstrom

metajack:
@Ms2ger Can you paste a link to a build log that shows this?

Ms2ger:
http://build.servo.org/builders/cargo-linux/builds/9/steps/cargo%20test/logs/stdio

PR #168, opened 2015-07-02
Make cargo test work

michaelwu:
Fixes https://github.com/servo/rust-mozjs/issues/161

### DIFF
--- a/src/rust.rs
+++ b/src/rust.rs
@@ -728,9 +728,11 @@ pub unsafe extern fn reportError(_cx: *mut JSContext, msg: *const c_char, report
 #[cfg(test)]
 pub mod test {
     use super::Runtime;
+    use jsapi::JS_Init;
 
     #[test]
     pub fn dummy() {
+        unsafe { assert_eq!(JS_Init(), 1); }
         let _rt = Runtime::new();
     }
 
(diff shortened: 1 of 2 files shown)


src/rust.rs
```diff
@@ -728,9 +728,11 @@ pub unsafe extern fn reportError(_cx: *mut JSContext, msg: *const c_char, report
 #[cfg(test)]
 pub mod test {
     use super::Runtime;
+    use jsapi::JS_Init;
 
     #[test]
     pub fn dummy() {
+        unsafe { JS_Init(); }
```
Ms2ger: Check the result, please.

Ms2ger:
I don't understand the linkhack change, but then again, I don't know if anybody does. Does Servo compile with this change?

src/rust.rs
michaelwu: We don't even do that in servo.

michaelwu:
Yeah, I checked on Linux.

michaelwu:
I mostly understand what's going on with the linkhack now. I just don't know how it worked before.

michaelwu:
Commit updated to use assert_eq, same as in servo.

PR #180, opened 2015-07-26
Add glue for header-inlined typed array functions that cause linker e…

jdm:
…rrors on OS X.

[Review on Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/180)


Ms2ger:
LGTM, but maybe @michaelwu has comments on ways to avoid this.

michaelwu:
There isn't any support for things in headers that aren't exported. Some inline functions are exported, but only as part of a class/struct definition that is exported. bindgen++ doesn't have support for non-exported things at the moment. I've had some thoughts about adding support for those things, but I haven't thought too deeply about it. At any rate, it's probably not going to be supported by the next bindings update, so this is fine.

metajack:
pinging @jdm 
Seems like this needs code changes.

PR #181, opened 2015-08-05
Remove symbols that are either missing or hidden in libmozjs

larsbergstrom:
There is one mangling-related removal (of va_args-related functions), but these
were never called and I'm not convinced they would have worked on any platform
except for Linux 64 bit.

r? @Ms2ger @michaelwu

[Review on Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/181)

### DIFF
--- a/src/jsapi.rs
+++ b/src/jsapi.rs
@@ -2760,18 +2760,8 @@ extern "C" {
 
 extern "C" {
     fn _Z9JS_AssertPKcS0_i(s: *const i8, file: *const i8, ln: i32);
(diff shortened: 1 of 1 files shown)


larsbergstrom:
These symbols only caused a problem on Android, where all of the symbols are force-resolved at startup. On other platforms, we have dynamic linking working, so the fact that these symbols don't exist doesn't cause a failure because we never attempt to call the Rust stub functions.

michaelwu:
r=me. This should be a useful reference when I try to fix this issue in bindgen++.

PR #187, opened 2015-08-13
Add `HeapSizeOf` implementation.

notriddle:
Part of servo/heapsize#5.

[Review on Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/187)


src/lib.rs
```diff
@@ -91,3 +96,12 @@ pub unsafe fn JS_ARGV(_cx: *mut JSContext, vp: *mut JSVal) -> *mut JSVal {
...
     *vp
 }
+
+// This is measured properly by the heap measurement implemented in SpiderMonkey.
+impl<T: Copy + GCMethods<T>> HeapSizeOf for Heap<T> {
+    fn heap_size_of_children(&self) -> usize {
+        0
+    }
+}
+known_heap_size!(0, JSVal);
```
jdm: This can be a derive annotation on the actual type now.

notriddle: The reason I didn't is that:

* UnsafeCell does not implement HeapSizeOf (and probably shouldn't, if we want to maintain aliasing guarantees). This means I can't derive HeapSizeOf Heap<T>.
* Why bring in a plugin for a one-line derive when a one-line macro will do? This means deriving JSVal is pointless.

jdm: I find the first argument more convincing, because deriving these things makes it much harder to make mistakes in the future if the internal representation of JSVal changes in the future.

notriddle: Okay. If I'm going to do that, then where is the `bindgen` program that `gen.py` calls? It seems to be generating the `jsapi::Value` struct.

jdm: That's https://github.com/michaelwu/rust-bindgen/ which also requires custom patches to clang among other things. We currently hand-edit these things. Anyways, I don't think it's a big deal at this point and we can move on.

PR #188, opened 2015-08-17
Bring back the enums that were lost in the rustup

frewsxcv:
We lost them in https://github.com/servo/rust-mozjs/pull/137

Related to https://github.com/servo/servo/issues/5273

[Review on Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/188)


SimonSapin:
They were replaced with consts because of https://github.com/rust-lang/rust/issues/23304

PR #192, opened 2015-08-20
Fix signature of JS_BufferIsCompilableUnit

tschneidereit:
r? @larsbergstrom

[Review on Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/192)

### DIFF
--- a/src/jsapi.rs
+++ b/src/jsapi.rs
@@ -4490,14 +4490,14 @@ extern "C" {
                                                                                              scopeChain:
                                                                                                  *mut AutoObjectVector)
      -> *mut JSObject;
-    fn _Z25JS_BufferIsCompilableUnitP9JSContextN2JS6HandleIP8JSObjectEEPKcj(cx:
+    fn _Z25JS_BufferIsCompilableUnitP9JSContextN2JS6HandleIP8JSObjectEEPKcm(cx:
                                                                                 *mut JSContext,
                                                                             obj:
                                                                                 Handle<*mut JSObject>,
                                                                             utf8:
                                                                                 *const i8,
                                                                             length:
-                                                                                u32)
+                                                                                u64)
      -> u8;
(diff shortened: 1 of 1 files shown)


src/jsapi.rs
```diff
@@ -4490,14 +4490,14 @@ extern "C" {
...
-    fn _Z25JS_BufferIsCompilableUnitP9JSContextN2JS6HandleIP8JSObjectEEPKcj(cx:
+    fn _Z25JS_BufferIsCompilableUnitP9JSContextN2JS6HandleIP8JSObjectEEPKcm(cx:
                                                                                 *mut JSContext,
                                                                             obj:
                                                                                 Handle<*mut JSObject>,
                                                                             utf8:
                                                                                 *const i8,
                                                                             length:
-                                                                                u32)
+                                                                                u64)
```
Manishearth: This is `size_t`. Should we be be using `usize`?

tschneidereit: >
> This is size_t. Should we be be using usize?
>
Probably, yes. It seems like bindgen generated wrong code for size_t? I
wonder if that afflicts more sites than just this one.

Manishearth: cc @michaelwu

Ms2ger: Yes and yes: #185

Manishearth: Is `libc::size_t` different from `usize`? It would be nice to use Rust ints whenever possible.

michaelwu:
https://gist.github.com/michaelwu/a9525f07112a8224a540 is the latest output from bindgen. `size_t` is handled correctly now.

Replacing `size_t` with usize sounds good - I'll try that next time, though I think we should stick to `size_t` for now for consistency.

michaelwu:
This should be fixed by the bindings update.

michaelwu:
https://github.com/servo/rust-mozjs/pull/198 will fix this.

PR #179, opened 2015-07-22
Replace `i8` with `libc::c_char`

akiss77:
It is definitely suboptimal to manually edit autogenerated code and
surely incomplete but, at least, gets servo build on platforms
where char is unsigned (e.g., on aarch64).

It would be great if a future version of bindgen could generate
`libc::c_char` at the right places.

[Review on Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/179)

PR #193, opened 2015-09-08
Manual fix to some function signatures to ensure correct usage of C characters

akiss77:
`libc::c_char` is not always `i8`, can be unsigned on some platforms
(e.g., on AArch64).

[Review on Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/193)


akiss77:
Similar to #179 
cc @michaelwu

PR #198, opened 2015-09-24
Regenerate bindings using the latest bindgen++

### LINKED ISSUE #158: Generate bindings per platform
Right now, there's a number of demangling functions which don't work since the mangled names contain types which are different between OSX/Linux or 32bit/64bit. Generating the bindings separately per platform will avoid this issue. This causes issues with the CEF port since the OSX linker wants all symbols to be defined.

CC @zmike

michaelwu:
Need to test this a bit more, but it's structured roughly the way I want it and no manual edits of jsapi_*.rs are required.

Fixes #158

[Review on Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/198)


michaelwu:
Tested and working on 64 bit OSX and 64 bit Linux. Had to fix an issue with the mangled names on OSX where an extra underscore was being prefixed to the mangled name.

michaelwu:
Tested and working on Gonk. Android is the only one left to check, I think.

michaelwu:
Adjusted the bindings generation a bit after @larsbergstrom found missing symbols on Android. Need to test once more, but I think this is ready for review. r? @jdm

michaelwu:
Works on Android.

jdm:
```
<jdm> but yeah, I'd be inclined to rewrite all of the generated transmutes to be pointer casts instead
<jdm> let raw = &mut self._bindgen_data_1_ as *mut _ as *mut u8; raw.offset(0) as *mut _
```

src/rust.rs
```diff
@@ -246,14 +271,14 @@ impl<T: Copy> DerefMut for MutableHandle<T> {
 
 impl HandleValue {
     pub fn null() -> HandleValue {
-        HandleValue {
-            ptr: &JSVAL_NULL
+        unsafe {
+            HandleValue::from_marked_location(&JSVAL_NULL)
         }
```
jdm: Let's use NullHandleValue.

src/rust.rs
```diff
@@ -246,14 +271,14 @@ impl<T: Copy> DerefMut for MutableHandle<T> {
...
     }
 
     pub fn undefined() -> HandleValue {
-        HandleValue {
-            ptr: &JSVAL_VOID
+        unsafe {
+            HandleValue::from_marked_location(&JSVAL_VOID)
         }
```
jdm: UndefinedHandleValue.

src/rust.rs
```diff
@@ -508,35 +543,41 @@ impl JSJitMethodCallArgs {
 //     to duplicate so much code here
 impl CallArgs {
     pub fn from_vp(vp: *mut Value, argc: u32) -> CallArgs {
+        use jsapi::{CallArgsBase, CallReceiverBase, IncludeUsedRval, UsedRvalBase};
```
jdm: I'd prefer this up with the other includes.

jdm:
```
* jdm is curious why FunctionClassPtr is a Class, while Int8ArrayClassPtr is a *const Class
 <mwu> jdm: looks like a bug. I think they should all be *const Class
<mwu> though.. this code is pretty clever, so I dunno
<mwu> probably a bug.
```

src/rust.rs
```diff
@@ -760,12 +801,12 @@ pub mod test {
...
             hasInstance: None,
             construct: None,
-            trace: Some(_Z24JS_GlobalObjectTraceHookP8JSTracerP8JSObject),
+            trace: Some(JS_GlobalObjectTraceHook),
 
             reserved: [0 as *mut libc::c_void; 25]
         };
 
-        unsafe { assert_eq!(JS_Init(), 1); }
+        unsafe { assert_eq!(JS_Init(), true); }
```
jdm: This could just be `unsafe { assert!(JS_Init()); }` now.

jdm:
Ok, I've pointed out all of the things in the generated output for linux32 that made me scratch my head, and the stuff in the non-generated files that could be improved, so I think my part here is complete.

PR #199, opened 2015-10-08
Check the result of JS_DefineFunction in the example.

Ms2ger:
[Review on Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/199)


jdm:
I think this will be easier to read as:
```rust
let func = JS_DefineFunction(...);
assert!(!func.is_null());
```

PR #206, opened 2015-10-23
Introduce define_methods and define_properties functions.

Ms2ger:
These are based on the functions of the same name in Servo's
script::dom::bindings::utils.

[Review on Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/206)

### DIFF
--- a/src/rust.rs
+++ b/src/rust.rs
@@ -34,11 +34,31 @@ use jsapi::{JSJitMethodCallArgs, JSJitGetterCallArgs, JSJitSetterCallArgs, CallA
 use jsapi::{NullHandleValue, UndefinedHandleValue, JSID_VOID};
 use jsapi::{CallArgsBase, CallReceiverBase, IncludeUsedRval, UsedRvalBase};
 use jsapi::CompartmentOptions;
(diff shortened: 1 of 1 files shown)


src/rust.rs
```diff
@@ -34,12 +34,26 @@ use jsapi::{JSJitMethodCallArgs, JSJitGetterCallArgs, JSJitSetterCallArgs, CallA
...
+use jsapi::JSNativeWrapper;
+use jsapi::JSPropertySpec;
+use jsapi::PropertyDefinitionBehavior;
 use jsval::UndefinedValue;
 use glue::{CreateAutoObjectVector, AppendToAutoObjectVector, DeleteAutoObjectVector};
 use glue::{NewCompileOptions, DeleteCompileOptions};
 use default_stacksize;
 use default_heapsize;
 
+fn ok(b: bool) -> Result<(), ()> {
```
jdm: I'd prefer a to_result method defined on bool; I find it really easy to misread `ok(something)` as `Ok(something)`.

Ms2ger: Fixed.

PR #208, opened 2015-10-29
Error reporting strings are Latin1, not UTF-8.

eefriedman:
Because the best way to convert from a 16-bit encoding to an 8-bit
encoding is obviously truncation.

[Review on Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/208)

### DIFF
--- a/src/rust.rs
+++ b/src/rust.rs
@@ -9,6 +9,7 @@ use libc::c_char;
 use std::char;
 use std::ffi;
 use std::ptr;
+use std::slice;
 use std::mem;
 use std::u32;
 use std::default::Default;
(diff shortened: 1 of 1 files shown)


michaelwu:
Have you tried using report->ucmessage?

eefriedman:
Just tried it... yes, it contains the exact same Latin1 string inflated to u16.

michaelwu:
Hmm I expected it to be the other way around. For example, in http://mxr.mozilla.org/mozilla-central/source/js/src/jscntxt.cpp#650 , ucmessage is the original string before truncation to latin1.

eefriedman:
Added commit to use ucmessage.  You can test for yourself if you want.

eefriedman:
Oops; not sure how I forgot it was in the standard library.  Updated.

eefriedman:
@michaelwu Ping.

eefriedman:
Nothing is happening here; do I need to file a bug against saltfs or something?

jdm:
@bors-servo: r=michaelwu
I had to re-synchronize homu's queue for rust-mozjs.

eefriedman:
The build timed out during the Spidermonkey build.  The configuration probably needs to be changed somehow.

jdm:
Bah, we're being bitten by libc here. #211 needs to merge first.

PR #210, opened 2015-11-03
Set nativeRegExp to true rather than false.

asajeffrey:
Improves SpiderMonkey regex performance, and gives us the same default as Firefox.

[Review on Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/210)


Ms2ger:
@bors-servo r+

Thank you for your PR.

PR #216, opened 2015-11-11
Make the `consts` module public

ecoal95:
[Review on Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/216)


ecoal95:
It's needed for https://github.com/servo/servo/pull/8412

r? @jdm (or anyone else really)

jdm:
Note that all the constants are still available through js:: instead of js::consts.

ecoal95:
Arrgh my fault... Sorry about that

PR #225, opened 2015-12-16
Add support for CustomAutoRooter

michaelwu:
As requested by @jdm . Fixes https://github.com/servo/servo/issues/7676. Not particularly great to use but it works for me.

[Review on Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/225)

### DIFF
--- a/src/rust.rs
+++ b/src/rust.rs
@@ -4,7 +4,7 @@
 
 //! Rust wrappers around the raw JS apis
 
-use libc::{size_t, c_uint, c_char};
+use libc::{size_t, c_uint, c_char, ptrdiff_t};
 use std::char;
 use std::ffi;
 use std::ptr;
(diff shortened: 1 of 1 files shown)


src/rust.rs
```diff
@@ -336,6 +337,40 @@ impl<T> Drop for Rooted<T> {
...
 
+impl CustomAutoRooter {
+    pub fn new(cx: *mut JSContext, vftable: &'static _vftable_CustomAutoRooter)
+        -> CustomAutoRooter {
+        CustomAutoRooter::new_with_addr(cx, vftable, unsafe { return_address() })
+    }
+
+    pub fn new_with_addr(cx: *mut JSContext, vftable: &'static _vftable_CustomAutoRooter, addr: *const u8) -> CustomAutoRooter {
+        let ctxfriend: &mut ContextFriendFields = unsafe {
+            mem::transmute(cx)
```
jdm: I would prefer `&mut *(cx as *mut ContextFriendFields)` instead of a transmute.

src/rust.rs
```diff
@@ -336,6 +337,40 @@ impl<T> Drop for Rooted<T> {
...
+        let rooter = CustomAutoRooter {
+            _vftable: vftable as *const _,
+            _base: AutoGCRooter {
+                down: ctxfriend.autoGCRooters,
+                tag_: AutoGCRooter_enum0::CUSTOM as ptrdiff_t,
+                stackTop: &mut ctxfriend.autoGCRooters as *mut _,
+            }
+        };
+
+        ctxfriend.autoGCRooters = unsafe { mem::transmute((addr as *const *const u8).offset(1)) };
```
jdm: I'd prefer explicit casts here instead of transmute.

src/rust.rs
```diff
@@ -336,6 +337,40 @@ impl<T> Drop for Rooted<T> {
...
+        rooter
+    }
+}
+
+impl Drop for CustomAutoRooter {
```
jdm: Bleh, another destructor on a repr(C) type :/

jdm:
I'd like to see the transmutes replaced with casts; otherwise this looks ok.

michaelwu:
Transmutes replaced with casts.

PR #226, opened 2015-12-23
conversions: Implement FromJSValConvertible for Vec<T>

ecoal95:
This will allow us to finally implement WebIDL sequence arguments
(servo/servo#544).

[Review on Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/226)

### DIFF
--- a/src/conversions.rs
+++ b/src/conversions.rs
@@ -30,11 +30,11 @@
 use JSPROP_ENUMERATE;
 use error::throw_type_error;
 use glue::RUST_JS_NumberValue;
-use jsapi::{JSContext, JSObject, JSString, Handle, HandleValue, MutableHandleValue};
+use jsapi::{JSContext, JSObject, JSString, HandleValue, MutableHandleValue};
 use jsapi::{JS_NewUCStringCopyN, JS_StringHasLatin1Chars, JS_WrapValue};
 use jsapi::{JS_GetLatin1StringCharsAndLength, JS_GetTwoByteStringCharsAndLength};
 use jsapi::{JS_NewArrayObject1, JS_DefineElement, RootedValue, RootedObject};
-use jsapi::{JS_GetArrayLength, JS_GetElement};
+use jsapi::{ForOfIterator, NonIterableBehavior};
 use jsval::{BooleanValue, Int32Value, NullValue, UInt32Value, UndefinedValue};
 use jsval::{JSVal, ObjectValue, ObjectOrNullValue, StringValue};
 use rust::{ToBoolean, ToNumber, ToUint16, ToInt32, ToUint32, ToInt64, ToUint64, ToString};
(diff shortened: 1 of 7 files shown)


Ms2ger:
This should implement the spec for "An ECMAScript value V is converted to an IDL sequence<T> value" at <https://heycam.github.io/webidl/#es-sequence>. SpiderMonkey has the [`ForOfIterator`](https://doc.servo.org/js/jsapi/struct.ForOfIterator.html) struct to help with that. You'll need to initialize it to match
```c++
    explicit ForOfIterator(JSContext* cx) : cx_(cx), iterator(cx_), index(NOT_ARRAY) { }
    static const uint32_t NOT_ARRAY = UINT32_MAX;
```

See <https://mxr.mozilla.org/mozilla-central/source/dom/bindings/Codegen.py?rev=151ce2b0e3f6#4514> for how Gecko calls it.

ecoal95:
@Ms2ger 

Added a minimal test case, that works if we remove the `Rooted<T>` destructor.

So... it seems that the hidden state `Drop` adds to `Rooted<T>` is what is corrupting our stack when interacting with SM, causing the trap an the latter assertion failure.

I don't think we can remove the Rooted destructor, so I don't think there's any way to circumvent this, at least for now.

Would it be ok to leave the first commit for now to land https://github.com/servo/servo/pull/9056 (so there's basic support for sequences and people can work on that kind of stuff), and leave this until we have a way to solve it using `ForOfIterator`?

If that sounds good, I'll make sure the corresponding issues are filled (both in here and in servo/servo).

I'd want to hear @jdm's opinion too.

ecoal95:
Also, pinging @michaelwu in case he knows any way to circumvent this issue.

Ms2ger:
Ah, crap. @tschneidereit, do you have thoughts on this?

Does the drop flag end up at the end or the start of the `Rooted`? If it's at the end, maybe a quick hack would be to move the fields of `ForOfIterator` around.

Ms2ger:
Alternatively, can't we opt out of the drop flag?

Manishearth:
`#[unsafe_no_drop_flag]`, however it's then up to you to maintain the drop flag invariants yourself

ecoal95:
I did a quick try moving the `iterator` to the end of `ForOfIterator` (both here and in mozjs) and no luck so far: it doesn't crash with `SIGTRAP`, but still crashes with our assertion in `Drop`.

Also tested with `#[unsafe_no_drop_flag]`, but it seems even in "simple" cases like that test we don't maintain the drop invariants and we drop many times, causing a segfault.

nox:
#227 should fix that.

ecoal95:
I just pushed a patch based on #227, but testing against the drop bitfield rust uses instead (see [comment](https://github.com/servo/rust-mozjs/pull/227#discussion_r48445511).

It works, but feels too hacky though. And we must touch the `jsapi_*` files manually too...

Ms2ger:
#229

PR #227, opened 2015-12-26
Remove the drop flags that can be removed

nox:
The remaining one is the one on `Heap<T>`. This should fix the issue related to #226.

Cc @ecoal95

[Review on Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/227)

### DIFF
--- a/src/rust.rs
+++ b/src/rust.rs
@@ -330,10 +330,14 @@ impl<T: Copy> MutableHandle<T> {
 
 impl<T> Drop for Rooted<T> {
     fn drop(&mut self) {
+        if self.stack.is_null() {
+            return;
+        }
         unsafe {
             assert!(*self.stack == mem::transmute(&*self));
             *self.stack = self.prev;
         }
+        self.stack = ptr::null_mut();
     }
 }
 
--- a/src/rust.rs
+++ b/src/rust.rs
@@ -370,10 +370,14 @@ impl CustomAutoRooter {
 
 impl Drop for CustomAutoRooter {
     fn drop(&mut self) {
+        if self._base.stackTop.is_null() {
+            return;
+        }
         unsafe {
             assert!(*self._base.stackTop == mem::transmute(&self._base));
             *self._base.stackTop = self._base.down;
         }
+        self._base.stackTop = ptr::null_mut();
     }
 }
 
--- a/src/rust.rs
+++ b/src/rust.rs
@@ -561,7 +561,11 @@ impl JSAutoRequest {
 
 impl Drop for JSAutoRequest {
     fn drop(&mut self) {
+        if self.mContext.is_null() {
+            return;
+        }
         unsafe { JS_EndRequest(self.mContext); }
+        self.mContext = ptr::null_mut();
     }
 }
 
--- a/src/rust.rs
+++ b/src/rust.rs
@@ -580,7 +580,11 @@ impl JSAutoCompartment {
 
 impl Drop for JSAutoCompartment {
     fn drop(&mut self) {
+        if self.cx_.is_null() {
+            return;
+        }
         unsafe { JS_LeaveCompartment(self.cx_, self.oldCompartment_); }
+        self.cx_ = ptr::null_mut();
     }
 }
 
--- a/src/jsapi_linux_32.rs
+++ b/src/jsapi_linux_32.rs
@@ -1452,6 +1452,7 @@ pub type JS_ICUFreeFn =
     ::std::option::Option<unsafe extern "C" fn(arg1: *const ::libc::c_void,
                                                p: *mut ::libc::c_void)>;
 #[repr(C)]
+#[unsafe_no_drop_flag]
 pub struct JSAutoRequest {
     pub mContext: *mut JSContext,
 }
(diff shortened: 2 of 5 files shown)


michaelwu:
Directly modifying the jsapi_*.rs files is generally not allowed - these files are fully autogenerated. You'll want to either make all generated structs have unsafe_no_drop_flag, add an annotation to mark certain structs as unsafe_no_drop_flag, or figure out some other way to convince rust not to add a drop flag (mark things as nonzero upstream and have bindgen wrap things in NonZero, maybe?).

nox:
@michaelwu Where should that be done? In your bindgen's fork? I suspected such a thing for starters, but I wanted to at least make a PR to know if that is actually a fix for #226.

nox:
On a related note, IMO that unsafe_no_drop_flag should actually live on the Drop impl.

michaelwu:
You'll most likely need hack bindgen. You may also want to add additional things to the upstream headers. I like the idea of wrapping things in NonZero if that'll make rust automatically remove the drop flag. https://github.com/servo/mozjs/pull/55 has the changes I made to the headers, so you can see some examples there.

src/rust.rs
```diff
@@ -330,10 +330,14 @@ impl<T: Copy> MutableHandle<T> {
 
 impl<T> Drop for Rooted<T> {
     fn drop(&mut self) {
+        if self.stack.is_null() {
```
ecoal95: Heh, I should have tried to write this. In my defence I was on a bus while testing.

I think this won't work, since rust fills the dropped objects (even those marked with `#[unsafe_no_drop_flag]`) with `0x1d` (see https://github.com/rust-lang/rust/blob/040a77f772d7693598499a161f53ed230fb61c52/src/libcore/mem.rs#L539).

A quick test with this patch applied over #227 yields:

```
(gdb) run
Starting program: /home/emilio/projects/moz/rust-mozjs/target/debug/js-d2b536494d16a8fc 
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".

running 2 tests
[New Thread 0x7ffff6b31700 (LWP 17770)]
[New Thread 0x7ffff5fff700 (LWP 17771)]
[New Thread 0x7ffff57ff700 (LWP 17772)]
[New Thread 0x7ffff4fff700 (LWP 17773)]
[New Thread 0x7ffff7fc0700 (LWP 17774)]
[New Thread 0x7ffff7f3f700 (LWP 17775)]
[New Thread 0x7ffff7ebe700 (LWP 17776)]
[New Thread 0x7ffff6930700 (LWP 17777)]
[New Thread 0x7ffff68af700 (LWP 17778)]
[New Thread 0x7ffff5dfe700 (LWP 17779)]
[New Thread 0x7ffff5d7d700 (LWP 17780)]
[New Thread 0x7ffff5cfc700 (LWP 17781)]
[New Thread 0x7ffff55fe700 (LWP 17782)]
[New Thread 0x7ffff557d700 (LWP 17783)]
[New Thread 0x7ffff54fc700 (LWP 17784)]
[New Thread 0x7ffff4dfe700 (LWP 17785)]

Program received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7ffff57ff700 (LWP 17772)]
0x000055555563bc7c in js::rust::Rooted<T>.Drop::drop (self=0x7ffff57fe980) at src/rust.rs:337
337	            assert!(*self.stack == mem::transmute(&*self));
(gdb) p self.stack
$1 = (struct Rooted<*mut libc::c_void> **) 0x1d1d1d1d1d1d1d1d
(gdb) p self
$2 = (struct Rooted<*mut js::jsapi_linux_64::JSObject> *) 0x7ffff57fe980
(gdb) p self._base
$3 = {_phantom0 = {<No data fields>}}
(gdb) p self.prev
$4 = (struct Rooted<*mut libc::c_void> *) 0x1d1d1d1d1d1d1d1d
(gdb) p self.ptr
$5 = (union JSObject *) 0x1d1d1d1d1d1d1d1d
```

We might want to check for `0x1d1d1d` (I just added a commit to #226 that does it, and works) , but it feels a bit too hacky...

Manishearth:
> I think this won't work, since rust fills the dropped objects (even those marked with #[unsafe_no_drop_flag]) with 0x1d

Wait, what?

The point of `unsafe_no_drop_flag` is that you're manually tracking the drop state somewhere, which can be inline. Won't the drop fill interact badly with it? I'd expect it not to happen for `unsafe_no_drop_flag`.

@asajeffrey was working with this attribute, maybe he knows how drop fill interacts with it.

ecoal95:
@Manishearth My fault, seems that the filling is done when contained in `ForOfIterator` (that in my experiment had no `#[unsafe_no_drop_flag]`. Adding that to `ForOfIterator` too in my patch would suffice, but we'll have to be careful when containing `Rooted`s in other structs.

Example: https://play.rust-lang.org/?gist=7436af45dc4ecf1d43f4&version=nightly

nox:
See https://github.com/michaelwu/rust-bindgen/pull/7 for another take at this.

nox:
@michaelwu What's blocking this now that it is rebased and that your fork of bindgen adds such drop flags?

michaelwu:
Does it work? Did we have to switch to checking for the poison instead of null? Also, I assume you've verified that bindgen does in fact remove the drop flag for these structs.

PR #234, opened 2016-01-17
Updated pub/priv

valarauca:
If you try to wrap the `<js::jsapi::Handle<T>>` class. You can't actually do this because the T type is platform specific, AND private. 

The example use case can be seen here: https://github.com/valarauca/lib_js/blob/feda0234c91901a606fcbccdef6b386374955a21/src/lib.rs#L21-L28

[Review on Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/234)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -24,19 +24,19 @@ extern crate rustc_serialize as serialize;
 
 #[cfg(target_os = "linux")]
 #[cfg(target_pointer_width = "64")]
-mod jsapi_linux_64;
+pub mod jsapi_linux_64;
 
 #[cfg(target_os = "macos")]
 #[cfg(target_pointer_width = "64")]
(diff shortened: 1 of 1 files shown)


ecoal95:
I think you can use `jsapi::JSObject` in that specific use case (or `HandleObject` directly).

Note that all public symbols from the platform-specific `jsapi` module are re-exported here: https://github.com/servo/rust-mozjs/blob/master/src/lib.rs#L41

valarauca:
Well that's my mistake sorry.

ecoal95:
Heh, no problem, I had a similar mistake not so long ago... https://github.com/servo/rust-mozjs/pull/216 :P 

Close this if you can please :-)

Ms2ger:
This works fine.

Issue #237, opened 2016-02-01
Conversion to vector should throw a TypeError 


Ms2ger:
Code is in `src/conversions.rs`, in `impl<C: Clone, T: FromJSValConvertible<Config=C>> FromJSValConvertible for Vec<T> {`.

Use `throw_type_error()` for the case where `is_object()` is false. The error should be something along the lines of "value is not an object".

stspyder:
@Ms2ger shouldn't we also throw a type error at https://github.com/servo/rust-mozjs/blob/master/src/conversions.rs#L517 for when a object without length property is passed on?

Ms2ger:
I don't remember when that case can be hit. We probably should for now.

jdm:
`new Blob({0: "foo"})`, I'm pretty sure.

ecoal95:
Don't we already throw from the generated code if all the possible conversions fail?

Ms2ger:
The contract for `from_jsval` says "If it returns `Err(())`, a JSAPI exception is pending." The handling in unions/overloads is broken, because we don't use for-of iteration, as we should. We'll have to figure out a way to do that correctly once the implementation here has been fixed.

KiChjang:
Fixed by #238.

PR #238, opened 2016-02-01
Throw type error on converting non objects to sequence

stspyder:
Fix for #237

[Review on Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/238)

### DIFF
--- a/src/conversions.rs
+++ b/src/conversions.rs
@@ -499,6 +499,7 @@ impl<C: Clone, T: FromJSValConvertible<Config=C>> FromJSValConvertible for Vec<T
         let mut length = 0;
 
         if !value.is_object() {
+            throw_type_error(cx, "Non objects cannot be converted to sequence");
             return Err(())
         }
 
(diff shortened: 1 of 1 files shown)


src/conversions.rs
```diff
@@ -508,7 +509,10 @@ impl<C: Clone, T: FromJSValConvertible<Config=C>> FromJSValConvertible for Vec<T
 
             for i in 0..length {
                 let mut val = RootedValue::new(cx, UndefinedValue());
-                assert!(JS_GetElement(cx, obj.handle(), i, val.handle_mut()));
+                if !JS_GetElement(cx, obj.handle(), i, val.handle_mut()) {
+                    // On JS exception or error exit loop
```
KiChjang: So no panic nor any error thrown?

stspyder: Doesn't 'JS_GetElement' automatically report JS exceptions? I thought so because the tests with assert_throws(custom_error, fun()) were passing.

KiChjang: Where is this assert_throws(custom_error, fun()) test?

KiChjang: As far as I can tell, JS_GetElement is an external C function, and it doesn't use any of the `throw_js_error` stuff that we have written in Rust.

stspyder: Well, I meant the WPT tests. I thought I saw the automatic exception reporting stuff somewhere in the SpiderMonkey documentation. May be it doesn't. I'll see if I can test.

Ms2ger: I thought this couldn't fail, but I double-checked, and it can:
```js
Object.defineProperty(Array.prototype, 1, {get: function() {throw 7}});
new Blob([,,,])
```

stspyder:
I have updated the pull request. From [JSAPI guide](https://developer.mozilla.org/en-US/docs/Mozilla/Projects/SpiderMonkey/JSAPI_User_Guide#Automatic_handling_of_uncaught_exceptions), it seems that JSAPI functions automatically sets the exception (thrown from JS) in the context and return false without we having to explicitly set one in the context. This exception will be cleared eventually down the stack.

I have updated the code to return an Err in case of a JS exception so that the constructor can return false if the conversion fails.

PR #249, opened 2016-03-29
Set the native stack quota per Gecko's defaults to defend against JavaScript stack overflows.

pcwalton:
Addresses servo/servo#10229.

r? @jdm

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/249)


jdm:
What about adding a test to src/tests.rs that initializes a basic global object, then evaluates `function f() { f() }` and asserts that the return value is 0?

pcwalton:
Sure, sounds like a good idea.

pcwalton:
Test added. r? @jdm

PR #258, opened 2016-04-22
Allow evaluate_script to return values

metajack:
This includes some other minor cleanups. This is a port of @tschneidereit's commits from this branch:
https://github.com/tschneidereit/rust-mozjs/tree/update-bindings

This is need to get https://github.com/tschneidereit/mozjs-shell  working.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/258)

### DIFF
--- a/src/rust.rs
+++ b/src/rust.rs
@@ -21,7 +21,7 @@ use jsapi;
 use jsapi::{JS_NewContext, JS_DestroyContext, JS_NewRuntime, JS_DestroyRuntime};
 use jsapi::{JSContext, JSRuntime, JSObject, JSFlatString, JSFunction, JSString, Symbol, JSScript, jsid, Value};
 use jsapi::{RuntimeOptionsRef, ContextOptionsRef, ReadOnlyCompileOptions};
-use jsapi::{JS_SetErrorReporter, Evaluate3, JSErrorReport};
+use jsapi::{JS_SetErrorReporter, Evaluate2, JSErrorReport};
 use jsapi::{JS_SetGCParameter, JSGCParamKey};
 use jsapi::{JSWhyMagic, Heap, Cell, HeapObjectPostBarrier, HeapValuePostBarrier};
 use jsapi::{ContextFriendFields};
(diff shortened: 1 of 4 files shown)


jdm:
The last commit will require nontrivial changes in Servo, I think. I'd rather have it separate from the rest of these changes.

metajack:
I'm not actually sure why it is need here. @till? I'm happy to remove it from this PR.

metajack:
Sorry @till. I meant @tschneidereit.

tschneidereit:
> The last commit will require nontrivial changes in Servo, I think. I'd rather have it separate from the rest of these changes.

It's not required. We should do the required changes to Servo anyway because long-term contexts will go away (and their uses be replaced by `JSRuntime`), and because there's no upside to using multiple contexts in Servo. But there's no need to do this now.

tschneidereit:
I _think_ only the first commit is needed. The second one makes sense, though, so we should keep it.

jdm:
@tschneidereit We only use one context. The commit in question is about requests.

tschneidereit:
> @tschneidereit We only use one context. The commit in question is about requests.

Hrm, sorry - it's been a while since I looked at all this and I misremembered stuff. What I really meant was that, AFAICT, there's no need to ever enter more than one request - or leave that request. We don't call `js::SetActivityCallback`, and I'm pretty sure that's the only thing triggered by `rt->requestDepth` being set to or changed from `0`.

Then again, it doesn't matter much at all whether we do or do not do this, so we can just ignore it.

Ms2ger:
We should call into jsapi with null for perf reasons, iirc; this PR doesn't change that, though.

nox:
Can we mark `evaluate_script` as unsafe? Especially if it doesn't use `JSAutoRequest ` even though it crashes without it.

Ms2ger:
Not entirely relatedly: maybe we should make `evaluate_script` a free function that takes `*mut JSContext`? The implementation shouldn't need a `Runtime`, and it can be annoying to get your hands on one.

metajack:
Rebased and dropped the last commit. Ready for rereview.

metajack:
@nox did you include this in your smup somehow? This hasn't landed but now mozjs-shell is working as expected, so I'm totally confused.

metajack:
oh, nm. github hadn't updated the page, and I was looking at stale data.

PR #263, opened 2016-05-13
Move Request handling into Runtime.

Ms2ger:
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/263)


nox:
* Are we sure this is safe?
* You didn't remove libc from Cargo.toml. Was that on purpose?

r=me after these two questions are addressed.

PR #264, opened 2016-05-18
Generalise Vec::<T>::from_jsval with ForOfIterator

nox:
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/264)


src/conversions.rs
```diff
@@ -495,30 +495,34 @@ impl<C: Clone, T: FromJSValConvertible<Config=C>> FromJSValConvertible for Vec<T
...
-        if !value.is_object() {
-            throw_type_error(cx, "Non objects cannot be converted to sequence");
+        let mut iterator = ForOfIterator {
+            cx_: cx,
+            iterator: RootedObject::new(cx, ptr::null_mut()),
+            index: ::std::u32::MAX, // NOT_ARRAY
+        };
+
+        if !iterator.init(value, NonIterableBehavior::AllowNonIterable) {
+            throw_type_error(cx, "Value is not iterable");
```
Ms2ger: No need to throw here, but you should if `!iterator.valueIsIterable()`

PR #265, opened 2016-05-23
Change Runtime::new to take a parent runtime

tschneidereit:
SpiderMonkey shares some data, such as the self-hosting compartment and read-only atoms, between runtimes - but only if a parent runtime is passed when calling `JS_NewRuntime`. This is important for both memory usage and performance: compiling the self-hosted JS takes more than 30ms on a desktop machine, so doing that every time, say, a worker is created is quite wasteful.

r? @nox

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/265)

### DIFF
--- a/src/rust.rs
+++ b/src/rust.rs
@@ -116,8 +116,8 @@ impl Runtime {
     ///
     /// Calling this function concurrently can cause segfaults inside
     /// SpiderMonkey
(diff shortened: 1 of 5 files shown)


nox:
Oh so that's why! Thanks for the explanation.

@bors-servo r+

PR #271, opened 2016-06-23
Fix linux 64bit bindings which included inline symbols

metajack:
This requires https://github.com/servo/rust-bindgen/pull/2 and https://github.com/servo/rust-bindgen/pull/3 to run bindgen, but does not need to wait on those to land as the new bindings are included here.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/271)


nox:
This looks ok.

@bors-servo r+

PR #272, opened 2016-06-26
Replace return_address usage in Rooted with a stack guard and a rooted! macro.

eddyb:
Part of a potential solution for rust-lang/rust#34227.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/272)

### DIFF
--- a/src/conversions.rs
+++ b/src/conversions.rs
@@ -33,7 +33,7 @@ use jsapi::JSPROP_ENUMERATE;
 use jsapi::{JSContext, JSObject, JSString, HandleValue, MutableHandleValue};
 use jsapi::{JS_NewUCStringCopyN, JS_StringHasLatin1Chars, JS_WrapValue};
 use jsapi::{JS_GetLatin1StringCharsAndLength, JS_GetTwoByteStringCharsAndLength};
-use jsapi::{JS_NewArrayObject1, JS_DefineElement, RootedValue, RootedObject};
+use jsapi::{JS_NewArrayObject1, JS_DefineElement, RootedObject};
 use jsapi::{ForOfIterator, ForOfIterator_NonIterableBehavior};
 use jsval::{BooleanValue, Int32Value, NullValue, UInt32Value, UndefinedValue};
 use jsval::{JSVal, ObjectValue, ObjectOrNullValue, StringValue};
@@ -473,10 +473,10 @@ impl<T: FromJSValConvertible> FromJSValConvertible for Option<T> {
 // https://heycam.github.io/webidl/#es-sequence
 impl<T: ToJSValConvertible> ToJSValConvertible for Vec<T> {
     unsafe fn to_jsval(&self, cx: *mut JSContext, rval: MutableHandleValue) {
-        let js_array = RootedObject::new(cx, JS_NewArrayObject1(cx, self.len() as libc::size_t));
+        rooted!(in(cx) let js_array = JS_NewArrayObject1(cx, self.len() as libc::size_t));
         assert!(!js_array.handle().is_null());
 
-        let mut val = RootedValue::new(cx, UndefinedValue());
+        rooted!(in(cx) let mut val = UndefinedValue());
         for (index, obj) in self.iter().enumerate() {
             obj.to_jsval(cx, val.handle_mut());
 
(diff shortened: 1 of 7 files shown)


src/rust.rs
```diff
@@ -269,39 +268,101 @@ impl RootKind for Value {
...
+    pub fn new(cx: *mut JSContext, root: &'a mut Rooted<T>) -> Self where T: RootKind {
+        unsafe {
+            root.add_to_root_stack(cx);
+        }
+        RootedGuard {
+            root: root
+        }
+    }
+
+    pub fn handle(&self) -> Handle<T> where T: Copy {
```
Ms2ger: Why the `Copy` bound here and on `handle_mut`?

Ms2ger:
Others should review this too, but I think this is good enough.

CC @nox @Manishearth @jdm

nox:
LGTM too.

src/conversions.rs
```diff
@@ -488,6 +488,29 @@ impl<T: ToJSValConvertible> ToJSValConvertible for Vec<T> {
     }
 }
 
+struct ForOfIteratorGuard<'a> {
```
Manishearth: documentation on this? (explaining why it is necessary perhaps)

Manishearth:
Overall lgtm. would like more docs.

src/rust.rs
eddyb: The methods to create those types requires it.

src/rust.rs
```diff
@@ -270,41 +268,103 @@ impl RootKind for Value {
...
-impl<T: RootKind + Copy> Rooted<T> {
-    pub fn new_with_addr(cx: *mut JSContext, initial: T, addr: *const u8) -> Rooted<T> {
-        let ctxfriend: &mut ContextFriendFields = unsafe {
-            mem::transmute(cx)
-        };
-
-        let kind = T::rootKind() as usize;
-        let root = Rooted::<T> {
+impl<T> Rooted<T> {
+    pub fn new_unrooted(initial: T) -> Rooted<T> {
```
fitzgen: My gut reaction to seeing an unrooted "`Rooted<T>`" is terror: it means you could pass around references to `Rooted<T>` (what would be `[Mutable]Handle<T>` in js/) that are not in fact rooted.

Is there any way we could use the type system to differentiate between things that are actually rooted vs those that are not (hashtag session types?), or maybe even just rename these types? It seems to me that `Rooted<T>` is really `Rootable<T>` and `RootedGuard<T>` is the real `Rooted<T>`.

eddyb: You cannot create a `Handle` safely from it. A reference to `Rooted` means nothing by itself.

src/rust.rs
```diff
@@ -270,41 +268,103 @@ impl RootKind for Value {
...
+        *self.stack = self.prev;
     }
+}
 
-    pub fn handle(&self) -> Handle<T> {
+/// Rust API for keeping a Rooted value in the context's root stack.
+/// Example usage: `rooted!(in(cx) let x = UndefinedValue());`.
+/// `RootedGuard::new` also works, but the macro is preferred.
+pub struct RootedGuard<'a, T: 'a> {
+    root: &'a mut Rooted<T>
```
fitzgen: The idea is that the `Rooted<T>` can safely be added to the root stack because it cannot be moved by the compiler while there is an active `&mut` reference to it?

src/rust.rs
eddyb: I would *love* to use better names but `Rooted` is an FFI type (which is required because it's a node in a linked list) so we cannot do too much about it.

src/rust.rs
eddyb: Yes, borrowed lvalues cannot be moved from in Rust. That's a core guarantee.

src/rust.rs
Ms2ger: Can we move the `initial` assignment to `RootedGuard::new()`?

eddyb: I suppose so? But what do you put there in `Rooted::new_unrooted`?

nox: [`GCMethods::initial`](https://doc.servo.org/js/rust/trait.GCMethods.html)

eddyb: That trait seems like a good choice to have as a bound in the struct itself.
Alright, other changes suggested on IRC were to have `struct Rootable<T>(Rooted<T>);` which could only be created empty and not accessed, and it's what `RootedGuard` would accept.

nox: You can do the wrapper that cannot be accessed and fill it with `initial`.

eddyb: However, that wrapper would cause issues with `ForOfIterator`, although that's internal to `rust-mozjs` and perhaps could be accommodated for with a `pub(crate)` unsafe method.

PR #276, opened 2016-07-08
Enable cross-crate inlining for JSVal conversion methods.

jdm:
Make it possible for Rust methods which are called in the generated bindings to be inlined. This shaved off 1/3-1/2 of the total time in the measurements in https://github.com/servo/servo/issues/12354.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/276)


emilio:
@bors-servo: r+

Doesn't rust inline this kind of methods cross-crate by default? Why?

jdm:
Cross-crate inline doesn't happen without the annotation as far as I know.

PR #277, opened 2016-07-13
More inlining possibilities

jdm:
These methods were showing up in profiles in https://github.com/servo/servo/issues/12354. This change reduced testcase time by 100ms.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/277)


jdm:
In general that's true. On the other hand, Gecko uses MOZ_ALWAYS_INLINE for stuff like this because we _really_ want these things to be inlined in the cases we care about.

PR #283, opened 2016-08-05
Fix up the .travis.yml file

larsbergstrom:
I think part of the reason we weren't getting builds is using super old syntax and bad `branches`.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/283)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,16 +7,12 @@ matrix:
   include:
     - os: linux
       env:
-        global:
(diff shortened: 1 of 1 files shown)


jdm:
What's the status here? The travis builds appear to be claiming that gcc 4.6.3 is available, but the build system requires 4.8 or newer.

larsbergstrom:
Stuff is working now; we were hitting some weird travis config issues when I originally created this, and I headed down a deep, dark path...

PR #284, opened 2016-08-05
Rewrite FromJSValConvertible to better support unions

### LINKED ISSUE #282: Rewrite FromJSValConvertible to better support unions
There are three cases to consider when converting a JS value to a Rust value:

1. Successful conversion
2. Impossible to convert the value
3. Exception is thrown from within the JS engine somewhere.

An example of case 3 happens when converting `{ get [Symbol.iterator]() { throw 7; } }` to a `sequence` type.

Right now, the signature is

```rust
unsafe fn from_jsval(cx: *mut JSContext,
                     val: HandleValue,
                     option: Self::Config)
                     -> Result<Self, ()>
```

and we use `throw_type_error` to set a pending exception in the second case.

That implies it is impossible for callers to detect whether case 2 or 3 occurred. This is fine for most cases, but unions need to be able to make that distinction (and *not* set a pending exception for the second case).

Possible solutions
==================
Just what I could come up with right now; feel free to propose other options.

Each section shows any `enum`s used in the proposal, followed by the common conversion case and the union case.

`Result<Self, Error>`
---------------------

```rust
enum Error {
    Case2(String),
    Case3
}

match T::from_jsval() {
    Ok(x) => x,
    Err(Case2(error)) => {
        throw_type_error(error);
        return Err(());
    }
    Err(Case3) => return Err(()),
}


match T::from_jsval() {
    Ok(x) => return Ok(Union::T(x)),
    Err(Case2(error)) => (),
    Err(Case3) => return Err(()),
}
```

`Result<Result<Self, String>, ()>`
----------------------------------

```rust
match try!(T::from_jsval()) {
    Ok(x) => x,
    Err(error) => {
        throw_type_error(error);
        return Err(());
    }
}

match try!(T::from_jsval()) {
    Ok(x) => return Ok(Union::T(x)),
    Err(error) => ()
}
```

`Result<ConversionResult<Self>, ()>`
------------------------------------

```rust
enum ConversionResult<T> {
    Success(T),
    Failure(String),
}

match try!(T::from_jsval()) {
    ConversionResult::Success(x) => x,
    ConversionResult::Failure(error) => {
        throw_type_error(error);
        return Err(());
    }
}

match try!(T::from_jsval()) {
    ConversionResult::Success(x) => return Ok(Union::T(x)),
    ConversionResult::Failure(error) => ()
}
```

`ConversionResult<Self>`
------------------------

```rust
enum ConversionResult<T> {
    Case1(T),
    Case2(String),
    Case3
}

match T::from_jsval() {
    Case1(x) => x,
    Case2(error) => {
        throw_type_error(error);
        return Err(());
    }
    Case3 => return Err(()),
}

match T::from_jsval() {
    Case1(x) => return Ok(Union::T(x)),
    Case2(error) => (),
    Case3 => return Err(()),
}
```

CC @nox @GuillaumeGomez

GuillaumeGomez:
Fixes #282.

This change is [Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/284)

### DIFF
--- a/src/conversions.rs
+++ b/src/conversions.rs
@@ -40,6 +40,7 @@ use jsval::{JSVal, ObjectValue, ObjectOrNullValue, StringValue};
 use rust::{ToBoolean, ToNumber, ToUint16, ToInt32, ToUint32, ToInt64, ToUint64, ToString};
 use libc;
 use num_traits::{Bounded, Zero};
+use std::borrow::Cow;
 use std::rc::Rc;
 use std::{ptr, slice};
 
@@ -91,6 +92,25 @@ pub trait ToJSValConvertible {
     unsafe fn to_jsval(&self, cx: *mut JSContext, rval: MutableHandleValue);
 }
 
+/// An enum to better support enums through FromJSValConvertible::from_jsval.
+#[derive(PartialEq, Eq, Clone, Debug)]
+pub enum ConversionResult<T> {
+    /// Everything went fine.
+    Success(T),
+    /// Pending exception.
+    Failure(Cow<'static, str>),
+}
+
+impl<T> ConversionResult<T> {
+    /// Returns Some(value) if it is `ConversionResult::Success`.
+    pub fn get_success_value(&self) -> Option<&T> {
+        match *self {
+            ConversionResult::Success(ref v) => Some(v),
+            _ => None,
+        }
+    }
+}
+
 /// A trait to convert `JSVal`s to Rust types.
 pub trait FromJSValConvertible: Sized {
     /// Optional configurable behaviour switch; use () for no configuration.
@@ -102,7 +122,7 @@ pub trait FromJSValConvertible: Sized {
     unsafe fn from_jsval(cx: *mut JSContext,
                          val: HandleValue,
                          option: Self::Config)
-                         -> Result<Self, ()>;
+                         -> Result<ConversionResult<Self>, ()>;
 }
 
 /// Behavior for converting out-of-range integers.
@@ -118,7 +138,7 @@ pub enum ConversionBehavior {
 
 /// Try to cast the number to a smaller type, but
 /// if it doesn't fit, it will return an error.
-unsafe fn enforce_range<D>(cx: *mut JSContext, d: f64) -> Result<D, ()>
+unsafe fn enforce_range<D>(cx: *mut JSContext, d: f64) -> Result<ConversionResult<D>, ()>
     where D: Bounded + As<f64>,
           f64: As<D>
 {
@@ -129,7 +149,7 @@ unsafe fn enforce_range<D>(cx: *mut JSContext, d: f64) -> Result<D, ()>
 
     let rounded = d.round();
     if D::min_value().cast() <= rounded && rounded <= D::max_value().cast() {
-        Ok(rounded.cast())
+        Ok(ConversionResult::Success(rounded.cast()))
     } else {
         throw_type_error(cx, "value out of range in an EnforceRange argument");
         Err(())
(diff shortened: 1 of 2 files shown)


GuillaumeGomez:
@Ms2ger: Tests work for me locally (on linux). Anything else I should check/modify to make it work on appveyor?

GuillaumeGomez:
Updated.

src/conversions.rs
```diff
@@ -481,12 +501,15 @@ impl<T: FromJSValConvertible> FromJSValConvertible for Option<T> {
     unsafe fn from_jsval(cx: *mut JSContext,
                          value: HandleValue,
                          option: T::Config)
-                         -> Result<Option<T>, ()> {
+                         -> Result<ConversionResult<Option<T>>, ()> {
         if value.get().is_null_or_undefined() {
-            Ok(None)
+            Ok(ConversionResult::Failure(Cow::Owned("Value is null or undefined.".to_owned())))
```
emilio: What's the point of returning a `Cow` if we only return owned values? Can't this be `Cow::Borrowed("Value is null or undefined")`?

GuillaumeGomez: cc @nox

emilio: @GuillaumeGomez: I'm pretty sure being able to return a borrowed cow is the reason for @nox to propose it, can you change it in the relevant places?

Ms2ger: This is not a failure case; should just stay `Ok(None)`.

src/conversions.rs
```diff
@@ -481,12 +501,15 @@ impl<T: FromJSValConvertible> FromJSValConvertible for Option<T> {
...
         } else {
-            let result: Result<T, ()> = FromJSValConvertible::from_jsval(cx, value, option);
-            result.map(Some)
+            match FromJSValConvertible::from_jsval(cx, value, option) {
```
Ms2ger: You can use `try!()` here

src/conversions.rs
```diff
@@ -553,7 +576,7 @@ impl<C: Clone, T: FromJSValConvertible<Config=C>> FromJSValConvertible for Vec<T
         let iterator = &mut *iterator.root;
 
         if !iterator.init(value, ForOfIterator_NonIterableBehavior::ThrowOnNonIterable) {
-            return Err(())
+            return Ok(ConversionResult::Failure(Cow::Owned("Not iterable".to_owned())))
```
Ms2ger: This doesn't agree with the `ThrowOnNonIterable` argument passed above. Please keep returning `Err(())`

Ms2ger:
`convert_int_from_jsval` and `enforce_range` need to be changed to return `Result<ConversionResult<D/T>, ()>`

GuillaumeGomez:
Updated.

src/conversions.rs
```diff
@@ -129,7 +149,7 @@ unsafe fn enforce_range<D>(cx: *mut JSContext, d: f64) -> Result<D, ()>
 
     let rounded = d.round();
     if D::min_value().cast() <= rounded && rounded <= D::max_value().cast() {
-        Ok(rounded.cast())
+        Ok(ConversionResult::Success(rounded.cast()))
     } else {
         throw_type_error(cx, "value out of range in an EnforceRange argument");
```
GuillaumeGomez: Should I return `Ok(ConversionResult::Failure)` here?

Ms2ger: Yep.

Ms2ger: Actually, no, I don't think so.

GuillaumeGomez: Ok.

PR #288, opened 2016-08-19
Allow a custom path to bindgen in etc/bindings.sh

fitzgen:
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/288)


Ms2ger:
Looks like we're getting to the point we should rewrite this script in python, but if it works, this is fine for now.

@bors-servo r+

PR #290, opened 2016-08-22
[do not merge] smup

fitzgen:
Matching PR for https://github.com/servo/mozjs/pull/93

Only has macos64 bindings thus far.

Is it correct to do

```
$ cargo build
$ ./etc/bindings.sh
$ mv out.rs src/jsapi_macos_64.rs
```

for the non-DEBUG bindings, and do

```
$ cargo build --features debugmozjs
$ ./etc/bindings.sh
$ mv out.rs src/jsapi_macos_64_debug.rs
```

for the DEBUG bindings? (Because that is what I have done here.)

@Ms2ger

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/290)

### DIFF
--- a/src/jsapi_macos_64.rs
+++ b/src/jsapi_macos_64.rs
@@ -863,9 +863,22 @@ pub struct _vftable_CallbackTracer {
 }
 #[repr(C)]
 #[derive(Debug, Copy)]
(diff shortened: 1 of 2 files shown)


Ms2ger:
#291

PR #300, opened 2016-08-29
Update etc/bindings.sh for the new bindgen flag format

fitzgen:
r? @emilio

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/300)

### DIFF
--- a/etc/bindings.sh
+++ b/etc/bindings.sh
@@ -3,11 +3,13 @@
 cd "$(dirname "$0")"
 
 EXTRA_FLAGS=
+EXTRA_CLANG_FLAGS=
 if [[ "$1" == "msvc14" ]] ; then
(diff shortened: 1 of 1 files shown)


emilio:
I'd have expected you to need to put `wrapper.h` after everything else, but if not, r=me

fitzgen:
> I'd have expected you to need to put wrapper.h after everything else, but if not, r=me

It is after all the normal flags, but before -- and the flags that get passed through to clang.

emilio:
@fitzgen: Yeah, I mean, for the `-x c++ -std=c++11` to be effective, I think you need to give the filename after it?

Anyway, I guess you've tested it, so feel free to invoke bors with `r=emilio`

PR #301, opened 2016-08-30
Implement maybe_wrap_object_value without OOL calls

nox:
The only remaining OOL calls are ToWindowIfWindowProxy and JS_WrapValue,
like in the equivalent Gecko function.

---

This change is [Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/301)

### DIFF
--- a/src/rust.rs
+++ b/src/rust.rs
@@ -16,46 +16,31 @@ use std::default::Default;
 use std::ops::{Deref, DerefMut};
 use std::cell::UnsafeCell;
 use std::marker::PhantomData;
-use consts::{JSCLASS_RESERVED_SLOTS_MASK, JSCLASS_GLOBAL_SLOT_COUNT, JSCLASS_IS_GLOBAL};
+use consts::{JSCLASS_RESERVED_SLOTS_MASK, JSCLASS_GLOBAL_SLOT_COUNT};
+use consts::{JSCLASS_IS_DOMJSCLASS, JSCLASS_IS_GLOBAL};
 use jsapi;
-use jsapi::{JS_Init, JS_GetContext, JS_NewRuntime, JS_DestroyRuntime};
-use jsapi::{JSContext, JSRuntime, JSObject, JSFlatString, JSFunction, JSString, Symbol, JSScript, jsid, Value};
-use jsapi::{RuntimeOptionsRef, ReadOnlyCompileOptions};
-use jsapi::{SetWarningReporter, Evaluate2, JSErrorReport};
-use jsapi::{JS_SetGCParameter, JSGCParamKey};
-use jsapi::{Heap, HeapObjectPostBarrier, HeapValuePostBarrier};
-use jsapi::{ContextFriendFields};
-use jsapi::{Rooted, Handle, MutableHandle, MutableHandleBase, RootedBase};
-use jsapi::{MutableHandleValue, HandleValue, HandleObject, HandleBase};
-use jsapi::AutoObjectVector;
-use jsapi::{ToBooleanSlow, ToNumberSlow, ToStringSlow};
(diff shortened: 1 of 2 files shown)


src/rust.rs
```diff
@@ -965,3 +955,79 @@ pub static SIMPLE_GLOBAL_CLASS: JSClass = JSClass {
...
+    assert!(rval.is_object_or_null());
+    if !rval.is_null() {
+        maybe_wrap_object_value(cx, rval);
+    }
+}
+
+#[inline]
+pub unsafe fn maybe_wrap_value(cx: *mut JSContext, rval: MutableHandleValue) {
+    if rval.is_object_or_null() {
+        maybe_wrap_object_value(cx, rval);
```
Ms2ger: `maybe_wrap_object_or_null_value`

src/rust.rs
```diff
@@ -965,3 +955,79 @@ pub static SIMPLE_GLOBAL_CLASS: JSClass = JSClass {
     cOps: &SIMPLE_GLOBAL_CLASS_OPS as *const JSClassOps,
     reserved: [0 as *mut _; 3]
 };
+
+#[inline]
+unsafe fn get_object_group(obj: *mut JSObject) -> *mut ObjectGroup {
+    assert!(!obj.is_null());
+    (*(obj as *mut Object)).group
```
Ms2ger: I think this would be more readable if you added a separate `let` binding for `obj as *mut Object`.

src/rust.rs
```diff
@@ -965,3 +955,79 @@ pub static SIMPLE_GLOBAL_CLASS: JSClass = JSClass {
...
+}
+
+#[inline]
+pub unsafe fn get_object_compartment(obj: *mut JSObject) -> *mut JSCompartment {
+    (*get_object_group(obj)).compartment
+}
+
+#[inline]
+pub unsafe fn get_context_compartment(cx: *mut JSContext) -> *mut JSCompartment {
+    (*(cx as *mut ContextFriendFields)).compartment_
```
Ms2ger: Ditto

nox:
Amended all the things.

Ms2ger:
Code looks okay to me; I'd like @jdm to agree we want to go this way, though.

(Also, some documentation might be useful.)

src/rust.rs
```diff
@@ -965,3 +955,81 @@ pub static SIMPLE_GLOBAL_CLASS: JSClass = JSClass {
...
+pub unsafe fn maybe_wrap_object_or_null_value(cx: *mut JSContext,
+                                              rval: MutableHandleValue) {
+    assert!(rval.is_object_or_null());
+    if !rval.is_null() {
+        maybe_wrap_object_value(cx, rval);
+    }
+}
+
+#[inline]
+pub unsafe fn maybe_wrap_value(cx: *mut JSContext, rval: MutableHandleValue) {
```
jdm: This is different than Gecko's, which special-cases strings and objects but otherwise avoids wrapping: https://dxr.mozilla.org/mozilla-central/rev/82d0a583a9a39bf0b0000bccbf6d5c9ec2596bcc/dom/bindings/BindingUtils.h#852

src/rust.rs
```diff
@@ -965,3 +955,81 @@ pub static SIMPLE_GLOBAL_CLASS: JSClass = JSClass {
...
+#[inline]
+pub unsafe fn maybe_wrap_object_value(cx: *mut JSContext, rval: MutableHandleValue) {
+    assert!(rval.is_object());
+    let obj = rval.to_object();
+    if get_object_compartment(obj) == get_context_compartment(cx) {
+        if is_dom_object(obj) {
+            return try_to_outerize(rval);
+        }
+    }
+    assert!(JS_WrapValue(cx, rval));
```
jdm: We miss a fast path here where the compartment is the same and it's a non-WebIDL object.

jdm:
Planning to finish this up, @nox?

PR #302, opened 2016-09-06
Add the `etc/bindings-all.py` script to orchestrate bindings generation

fitzgen:
The `bindings-all.py` script will make sure that you don't accidentally try and use DEBUG bindings for a non-DEBUG build or vice-versa. It orchestrates all of building, regenerating and testing new SpiderMonkey bindings.

Running this script on fedora regenerates the linux_64 bindings exactly as they are on the smup branch for me. I think this is a large usability improvement over what we do now.

r? @Ms2ger @nox

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/302)

### DIFF
--- a/etc/bindings-all.py
+++ b/etc/bindings-all.py
@@ -1,0 +1,82 @@
+#!/usr/bin/env python3
+
+import os, subprocess, sys
+
+def usage_and_exit():
+    print("")
+    print("Usage:")
+    print("    bindings-all.py <platform> <bindgen> <clang_lib_path>")
+    print("")
+    print("    Regenerates both debugmozjs and non-debugmozjs bindings and")
+    print("    tests them.")
+    print("")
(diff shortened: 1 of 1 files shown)


etc/bindings-all.py
```diff
@@ -0,0 +1,90 @@
...
+    ("_debug", ["--features", "debugmozjs"])
+]
+
+for (build_modifier, extra_cargo_flags) in BUILDS:
+    run(["cargo", "clean"])
+
+    try:
+        run(["cargo", "build"] + extra_cargo_flags)
+    except subprocess.CalledProcessError:
+        # This is expected to initially fail because the bindings aren't up to
```
Ms2ger: Can you try making this call `cargo build -p mozjs_sys` instead? r+ either way.

fitzgen: Fixed and force pushed

PR #304, opened 2016-09-13
Improve typed array support.

Ms2ger:
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/304)

### DIFF
--- a/src/typedarray.rs
+++ b/src/typedarray.rs
@@ -1,0 +1,305 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+//! High-level, safe bindings for JS typed array APIs. Allows creating new
+//! typed arrays or wrapping existing JS reflectors, and prevents reinterpreting
+//! existing buffers as different types except in well-defined cases.
+
+use glue::GetFloat32ArrayLengthAndData;
+use glue::GetFloat64ArrayLengthAndData;
+use glue::GetInt16ArrayLengthAndData;
+use glue::GetInt32ArrayLengthAndData;
+use glue::GetInt8ArrayLengthAndData;
+use glue::GetUint16ArrayLengthAndData;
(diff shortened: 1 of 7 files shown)


jdm:
I am super pumped to look at this later today!

src/typedarray.rs
```diff
@@ -0,0 +1,314 @@
...
+
+/// Internal trait for creating new typed arrays.
+pub trait TypedArrayElementCreator: TypedArrayElement {
+    /// Create a new typed array.
+    unsafe fn create_new(cx: *mut JSContext, length: u32) -> *mut JSObject;
+    /// Get the data.
+    unsafe fn get_data(obj: *mut JSObject) -> *mut Self::Element;
+}
+
+macro_rules! typed_array_element {
```
jdm: I propose one macro branch that accepts $t, $element, $unwrap, and $length_and_data, which declares the struct and implements TypedArrayElement, and another branch that accepts all the existing arguments, invokes the previous branch, and implements TypedArrayElementCreator.

src/typedarray.rs
```diff
@@ -0,0 +1,314 @@
...
+                     UnwrapArrayBuffer,
+                     GetArrayBufferLengthAndData,
+                     JS_NewArrayBuffer,
+                     JS_GetArrayBufferData);
+
+
+/// The ArrayBufferView type.
+pub struct ArrayBufferViewU8;
+
+impl TypedArrayElement for ArrayBufferViewU8 {
```
jdm: We shouldn't need this if we modify the typed_array_element macro as suggested above.

jdm:
As for the neutering stuff, I've spent a while looking at Gecko's implementation and thinking about whether we can do anything better, and I haven't come up with any good ideas. APIs like TypedArray::with_slice could accept a closure, but that wouldn't prevent people from running JS that could invalidate the slice inside of the closure, so it would still be unsafe and less pleasant to use.

Ms2ger:
Still needs tests.

Ms2ger:
Tests done, r? @jdm

jdm:
@bors-servo: r+
Thanks!

PR #305, opened 2016-09-15
Add more APIs to work with AutoIdVector.

Ms2ger:
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/305)

### DIFF
--- a/src/rust.rs
+++ b/src/rust.rs
@@ -47,6 +47,8 @@ use jsval::UndefinedValue;
 use jsapi::JSCLASS_RESERVED_SLOTS_SHIFT;
 use jsapi::JSClassOps;
 use jsapi::InitSelfHostedCode;
+use jsapi::AutoIdVector;
+use glue::{CreateAutoIdVector, SliceAutoIdVector, DestroyAutoIdVector};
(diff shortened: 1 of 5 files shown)


Ms2ger:
This is working towards https://github.com/servo/servo/issues/13144

src/rust.rs
```diff
@@ -882,6 +884,48 @@ impl JSNativeWrapper {
...
+impl IdVector {
+    pub fn new(cx: *mut JSContext) -> IdVector {
+        unsafe {
+            let vector = CreateAutoIdVector(cx);
+            assert!(!vector.is_null());
+            IdVector(vector)
+        }
+    }
+
+    pub unsafe fn get(&self) -> *mut AutoIdVector {
```
nox: This can be safe.

src/rust.rs
```diff
@@ -882,6 +884,48 @@ impl JSNativeWrapper {
...
+        self.0
+    }
+
+    pub fn append(&mut self, id: jsid) -> bool {
```
nox: This should be unsafe, given we can pass arbitrary `jsid` values.

PR #306, opened 2016-09-19
Return a ConversionResult::Failure when converting a non-iterable value to Vec.

Ms2ger:
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/306)


Ms2ger:
This is working towards servo/servo#13144.

PR #309, opened 2016-10-25
Handle panics.

Ms2ger:
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/309)

### DIFF
--- a/src/panic.rs
+++ b/src/panic.rs
@@ -1,0 +1,33 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+use std::any::Any;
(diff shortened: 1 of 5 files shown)


src/panic.rs
```diff
@@ -0,0 +1,33 @@
...
+/// Generic wrapper for JS engine callbacks panic-catching
+pub fn wrap_panic<F, R>(function: F, generic_return_type: R) -> R
+    where F: FnMut() -> R + UnwindSafe
+{
+    let result = catch_unwind(function);
+    match result {
+        Ok(result) => result,
+        Err(error) => {
+            PANIC_RESULT.with(|result| {
+                assert!(result.borrow().is_none());
```
jdm: When we hit this assertion in https://github.com/servo/servo/issues/13635 we get no information about either the existing stored panic or the new one. Can we improve that somehow?

jdm:
@bors-servo: r+
We can fix this code later.

PR #311, opened 2016-11-02
Fix code involving empty type references

DemiMarie:
by changing it to use pointers instead.

Fixes #311

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/311)


Ms2ger:
@bors-servo r+

Thanks!

PR #314, opened 2016-11-10
Implement a Trace trait for types that can be traced.

ejpbruel:
We expose several JSAPI functions that are used for tracing. By abstracting these functions behind
a trait we can overload them on the type to be traced.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/314)


ejpbruel:
cc @jimblandy

PR #319, opened 2016-11-21
Add methods to create a HandleValueArray.

ejpbruel:
Rust-mozjs currently does not define any methods to create instances of HandleValueArray.
Consequently, consumers have to create these instances manually. This is unnecessary boilerplate,
and therefore should be abstracted behind a function.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/319)

### DIFF
--- a/src/rust.rs
+++ b/src/rust.rs
@@ -20,8 +20,8 @@ use consts::{JSCLASS_RESERVED_SLOTS_MASK, JSCLASS_GLOBAL_SLOT_COUNT};
 use consts::{JSCLASS_IS_DOMJSCLASS, JSCLASS_IS_GLOBAL};
 use jsapi;
 use jsapi::{AutoIdVector, AutoObjectVector, CallArgs, CompartmentOptions, ContextFriendFields};
-use jsapi::{Evaluate2, Handle, HandleBase, HandleObject, HandleValue, Heap, HeapObjectPostBarrier};
-use jsapi::{HeapValuePostBarrier, InitSelfHostedCode, IsWindowSlow, JS_BeginRequest};
+use jsapi::{Evaluate2, Handle, HandleBase, HandleObject, HandleValue, HandleValueArray, Heap};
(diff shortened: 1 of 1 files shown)


src/rust.rs
```diff
@@ -495,6 +495,22 @@ impl HandleValue {
...
+        HandleValueArray {
+            length_: 0,
+            elements_: ptr::null_mut(),
+        }
+    }
+
+    pub fn from_slice(values: &[Value]) -> HandleValueArray {
+        HandleValueArray {
+            length_: values.len(),
+            elements_: values.as_ptr()
```
jdm: How does this compile?  as_ptr doesn't return a mutable pointer.

src/rust.rs
```diff
@@ -495,6 +495,22 @@ impl HandleValue {
     }
 }
 
+impl HandleValueArray {
+    pub fn new() -> HandleValueArray {
+        HandleValueArray {
+            length_: 0,
+            elements_: ptr::null_mut(),
```
jdm: This should be `ptr::null()` instead.

src/rust.rs
```diff
@@ -495,6 +495,22 @@ impl HandleValue {
...
+        }
+    }
+
+    pub fn from_slice(values: &[Value]) -> HandleValueArray {
```
jdm: This is not clearly safe to me - it's easy to create a vector which isn't rooted, leading to a handle that can point to things that get GCed. We should either remove this constructor, figure out a way to make it safe, or mark it unsafe.

ejpbruel: Fair enough. Having this constructor is something I depend on quite often in the Rust Debugger API, so I'd prefer to keep it around.

I did try to make this function take a &[HandleValue] instead, since that would guarantee the values are rooted. Unfortunately, a HandleValue is not a pointer, but rather a pointer to a pointer, so you can't use it to create a HandleValueArray directly. Nor can we create a Vec<Value> from the &[HandleValue], since there's no way to keep that alive after this function returns.

Given the above, I'd like to opt to make this function unsafe.

jdm: Unsafe and called `from_rooted_slice` sounds ok to me.

PR #320, opened 2016-11-22
Implement `AsHandle` and `AsHandleMut` traits.

ejpbruel:
Several types in rust-mozjs, such as Heap, PersistentRooted, and RootedGuard, represent a rooted
    value. A common property of rooted values is that a handle can be obtained to it. Rather than
    implement this property separately for each of these types, we should abstract it out into these
    two traits.
    
    This has several advantages: for instance, one can write functions that take any rooted value as
    argument, regardless of how it is implemented, and then obtain a handle to it. Additionally, it
    allows one to implement these traits for newtypes which underlying type is a rooted value.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/320)

### DIFF
--- a/src/rust.rs
+++ b/src/rust.rs
@@ -20,8 +20,8 @@ use consts::{JSCLASS_RESERVED_SLOTS_MASK, JSCLASS_GLOBAL_SLOT_COUNT};
 use consts::{JSCLASS_IS_DOMJSCLASS, JSCLASS_IS_GLOBAL};
 use jsapi;
 use jsapi::{AutoIdVector, AutoObjectVector, CallArgs, CompartmentOptions, ContextFriendFields};
-use jsapi::{Evaluate2, Handle, HandleBase, HandleObject, HandleValue, Heap, HeapObjectPostBarrier};
-use jsapi::{HeapValuePostBarrier, InitSelfHostedCode, IsWindowSlow, JS_BeginRequest};
+use jsapi::{Evaluate2, Handle, HandleBase, HandleObject, HandleValue, HandleValueArray, Heap};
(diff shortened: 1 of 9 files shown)


src/rust.rs
```diff
@@ -596,28 +637,9 @@ impl<T: GCMethods<T> + Copy> Heap<T> {
         ptr
     }
 
-    pub fn set(&mut self, v: T) {
-        unsafe {
-            let ptr = self.ptr.get();
-            let prev = *ptr;
-            *ptr = v;
-            T::post_barrier(ptr, prev, v);
```
jdm: Why is it safe to avoid this post barrier?

src/rust.rs
```diff
@@ -320,6 +321,50 @@ unsafe impl Trace for Heap<jsid> {
...
+
+    fn set(&mut self, v: T) where T: Copy {
+        self.handle_mut().set(v)
+    }
+}
+
+/// Given the types `$T` and `$U`, implements the trait `AsHandleMut<$T>` for the type `$U`, where
+/// `$U` is a newtype whose underlying type implements `HandleMut<$T>`.
+#[macro_export]
+macro_rules! derive_as_handle_mut {
```
jdm: I don't really see a purpose for defining these macros here. The implementations aren't hard to write by hand, and I haven't come up with a use case for this in Servo yet.

jimblandy:
I suggested `AsHandle` and `AsHandleMut` because they would help write generic code that could take a reference to any value that could provide a handle, rather than making the callers extract the handles themselves. It seems analogous to the way the `std::fs` functions accept any argument that implements `AsRef<Path>`.

jdm:
Closing due to lack of activity.

PR #321, opened 2016-12-12
Implement FromJSValConvertible for *mut JSObject

KiChjang:
r? @Ms2ger or @nox

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/321)

### DIFF
--- a/src/conversions.rs
+++ b/src/conversions.rs
@@ -623,10 +623,34 @@ impl ToJSValConvertible for *mut JSObject {
 }
 
 // https://heycam.github.io/webidl/#es-object
+impl FromJSValConvertible for *mut JSObject {
+    type Config = ();
+    #[inline]
+    unsafe fn from_jsval(_cx: *mut JSContext,
+                         val: HandleValue,
+                         _option: ())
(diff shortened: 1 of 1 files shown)


Ms2ger:
Doesn't compile

KiChjang:
Uhh right, stupid me.

Ms2ger:
These implementations crash if handed the wrong type of value. Is that intentional?

KiChjang:
I essentially copied the implementation from JSVal. Shall I instead check that it is an object first, and if not, return an Err(()) or a ConversionResult::Failure?

Ms2ger:
Let's take a step back. Why did you make this PR?

KiChjang:
MessagePort contains a WebIDL method that looks like this:
```
  void postMessage(any message, optional sequence<object> transfer = []);
```
And so CodegenRust.py is trying to convert a JS HandleValue to a JS object, which currently doesn't exist in rust-mozjs bindings.

jdm:
Given that this is only needed because of codegen for a feature we don't support yet (transferrable objects), I propose we use `optional any transfer` and close this PR.

PR #322, opened 2016-12-20
Store the JSContext in TLS for easy access.

Ms2ger:
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/322)

### DIFF
--- a/src/rust.rs
+++ b/src/rust.rs
@@ -14,7 +14,7 @@ use std::mem;
 use std::u32;
 use std::default::Default;
 use std::ops::{Deref, DerefMut};
-use std::cell::UnsafeCell;
+use std::cell::{Cell, UnsafeCell};
 use std::marker::PhantomData;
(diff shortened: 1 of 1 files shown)


nox:
Why not put the TLS thing in servo/servo? Why forbid the creation of two unrelated runtimes on a single thread?

fitzgen:
> Why forbid the creation of two unrelated runtimes on a single thread?

SpiderMonkey requires a 1:1 relationship between a `JSRuntime` and its thread.

nox:
@fitzgen Thanks!

@Ms2ger Could you make `::get()` panic if the pointer is null?

nox:
I still believe that was the wrong way to go, because now if we want to use the `Runtime` value and `Runtime::get` to do safe abstractions around for example `ToJsValConvertible`, we still won't be able to avoid some unsafe code in servo for dereferencing the result of `Runtime::get`, whereas we could make it be owned by TLS if it were done in servo instead of here.

PR #326, opened 2017-01-13
Make Rooted do its job.

Ms2ger:
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/326)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,8 @@ name = "evaluate"
 [[test]]
 name = "panic"
 [[test]]
+name = "rooting"
+[[test]]
 name = "typedarray"
 [[test]]
 name = "stack_limit"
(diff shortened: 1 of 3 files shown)


Ms2ger:
My main fear is that the two out-of-line calls will be terrible for performance. I don't suppose we use `Rooted` terribly often…

Ms2ger:
@jdm can you check the fixup commit? I'm not sure if the additional unsafety is worth it.

jdm:
As mentioned on IRC, I think we should follow jonco's suggestion from https://github.com/servo/servo/issues/13096#issuecomment-250901349 rather than pursue the patch I wrote in that issue.

jdm:
I'm glad to see the test, though!

jdm:
@bors-servo: r+
Nice.

PR #327, opened 2017-01-18
Ensure Runtime::get() works during the final GC under JS_DestroyRuntime.

Ms2ger:
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/327)


jdm:
A test verifying this behaviour would be ideal, whether in this repository of as a web-platform-test.

PR #333, opened 2017-02-04
Move code from TypedArray::create and Add TypedArray::update

deror1869107:
For servo/servo#15350

```TypedArray::update``` is used to be replace ```update_array_buffer_view```

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/333)

### DIFF
--- a/src/typedarray.rs
+++ b/src/typedarray.rs
@@ -17,6 +17,7 @@ use glue::GetUint8ArrayLengthAndData;
 use glue::GetUint8ClampedArrayLengthAndData;
 use jsapi::GetArrayBufferLengthAndData;
 use jsapi::GetArrayBufferViewLengthAndData;
+use jsapi::HandleObject;
 use jsapi::JSContext;
 use jsapi::JSObject;
 use jsapi::JS_GetArrayBufferData;
@@ -136,12 +137,23 @@ impl<'a, T: TypedArrayElementCreator + TypedArrayElement> TypedArray<'a, T> {
         }
 
         if let Some(data) = data {
-            assert!(data.len() <= length as usize);
-            let buf = T::get_data(result.get());
-            ptr::copy_nonoverlapping(data.as_ptr(), buf, data.len());
+            TypedArray::<T>::update_raw(data, result.handle());
         }
 
         Ok(())
+    }
+    
+    ///  Update an existed JS typed array
+    pub unsafe fn update(&mut self,
+                         data: &[T::Element]) {
+        TypedArray::<T>::update_raw(data, self.object.handle());
+    }
+
+    unsafe fn update_raw(data: &[T::Element],
+                         result: HandleObject) {
+        let (buf, length) = T::length_and_data(result.get());
+        assert!(data.len() <= length as usize);
+        ptr::copy_nonoverlapping(data.as_ptr(), buf, data.len());
     }
 }
 
--- a/tests/typedarray.rs
+++ b/tests/typedarray.rs
@@ -46,8 +46,33 @@ fn typedarray() {
         typedarray!(in(cx) let array: Uint32Array = rval.get());
         assert_eq!(array.unwrap().as_slice(), &[1, 3, 5, 0, 0][..]);
 
+        typedarray!(in(cx) let mut array: Uint32Array = rval.get());
+        array.as_mut().unwrap().update(&[0, 2, 4, 6]);
+        assert_eq!(array.unwrap().as_slice(), &[0, 2, 4, 6, 0][..]);
+
         rooted!(in(cx) let rval = ptr::null_mut());
         typedarray!(in(cx) let array: Uint8Array = rval.get());
         assert!(array.is_err());
+    }
+}
+
+#[test]
+#[should_panic]
+fn typedarray_update_panic() {
+    let rt = Runtime_::new();
+    let cx = rt.cx();
+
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: src/typedarray.rs [hunk 1]
```diff
@@ -132,13 +132,20 @@ impl<'a, T: TypedArrayElementCreator + TypedArrayElement> TypedArray<'a, T> {
...
 
+        TypedArray::<T>::update(length, data, result);
+        
+        Ok(())
+    }
+    
+    ///  Update an existed JS typed array
+    pub unsafe fn update(length: u32,
+                         data: Option<&[T::Element]>,
+                         result: MutableHandleObject) {
```
jdm: Rather than making a static method like this, I recommend `pub fn update(&mut self, data: &[T::Element])`.

deror1869107: But fn ```create```  and ```update``` don't use any field of the ```struct TypedArray```, why?

deror1869107: However if it isn't a static method, I can't call it from ```TypedArray::create```.

jdm: I argue that we can create a update_raw that encapsulates the relevant bit, and call that from both create and update (passing in a mutable destination buffer and an immutable source buffer).

deror1869107: Should I replace all ```data: Option<&[T::Element]>``` with ```data: <&[T::Element]>```?

deror1869107:
Should I replace all ```data: Option<&[T::Element]>``` with ```data: <&[T::Element]>```?

jdm:
Yes, that would lead to a cleaner API.

src/typedarray.rs
```diff
@@ -124,22 +124,31 @@ impl<'a, T: TypedArrayElementCreator + TypedArrayElement> TypedArray<'a, T> {
     /// be copied into the newly-allocated buffer. Returns the new JS reflector.
     pub unsafe fn create(cx: *mut JSContext,
                          length: u32,
-                         data: Option<&[T::Element]>,
```
jdm: We want to keep the Option here, and only call update_raw if it's Some.

src/typedarray.rs
```diff
@@ -124,22 +124,31 @@ impl<'a, T: TypedArrayElementCreator + TypedArrayElement> TypedArray<'a, T> {
...
         Ok(())
     }
+    
+    ///  Update an existed JS typed array
+    pub unsafe fn update(&mut self,
+                         data: &[T::Element]) {
+        TypedArray::<T>::update_raw(data, self.object.handle_mut());
+    }
+
+    pub unsafe fn update_raw(data: &[T::Element],
```
jdm: This should not be public.

src/typedarray.rs
```diff
@@ -124,22 +124,31 @@ impl<'a, T: TypedArrayElementCreator + TypedArrayElement> TypedArray<'a, T> {
...
+                             result: MutableHandleObject) {
```
jdm: This can be a `HandleObject` instead.

tests/typedarray.rs
```diff
@@ -45,5 +45,9 @@ fn typedarray() {
 
         typedarray!(in(cx) let array: Uint32Array = rval.get());
         assert_eq!(array.unwrap().as_slice(), &[1, 3, 5, 0, 0][..]);
+
+        typedarray!(in(cx) let mut array: Uint32Array = rval.get());
+        array.as_mut().unwrap().update(Some(&[0, 2, 4, 6]));
+        assert_eq!(array.unwrap().as_slice(), &[0, 2, 4, 6, 0][..]);
     }
```
jdm: Let's add another test in this file that uses `#[should_panic]` that calls update with an argument that is longer than the typed array that it is replacing, too.

jdm:
Not sure why my last comment is marked outdated, but please look at it as well.

deror1869107:
@jdm

About panic test? I have added it already!

src/typedarray.rs
```diff
@@ -132,14 +133,25 @@ impl<'a, T: TypedArrayElementCreator + TypedArrayElement> TypedArray<'a, T> {
...
 
-        if let Some(data) = data {
-            assert!(data.len() <= length as usize);
-            let buf = T::get_data(result.get());
-            ptr::copy_nonoverlapping(data.as_ptr(), buf, data.len());
+        if data.is_some() {
+            TypedArray::<T>::update_raw(data.unwrap(), result.handle());
         }
-
+
```
jdm: nit: remove the trailing spaces on this line.

src/typedarray.rs
```diff
@@ -132,14 +133,25 @@ impl<'a, T: TypedArrayElementCreator + TypedArrayElement> TypedArray<'a, T> {
             return Err(());
         }
 
-        if let Some(data) = data {
-            assert!(data.len() <= length as usize);
-            let buf = T::get_data(result.get());
-            ptr::copy_nonoverlapping(data.as_ptr(), buf, data.len());
+        if data.is_some() {
+            TypedArray::<T>::update_raw(data.unwrap(), result.handle());
```
jdm: We can use `if let` here instead and avoid the need to unwrap the value.

deror1869107:
Conflicts resolved

jdm:
@bors-servo: r+
Thanks!

PR #334, opened 2017-02-06
Protect TypedArray::from method against null pointers

### LINKED ISSUE #332: TypedArray::from method should protect against null pointers
We should add a testcase to `tests/typedarray.rs` that uses the `typedarray!` macro on `ptr::null_mut()`, and write code that correctly returns an Err value if that's encountered.

Code: `src/typedarray.rs`

zakorgy:
Fixes #332. Also added a test as described in the issue.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/334)

### DIFF
--- a/src/typedarray.rs
+++ b/src/typedarray.rs
@@ -70,6 +70,9 @@ impl<'a, T: TypedArrayElement> TypedArray<'a, T> {
                 root: &'a mut Rooted<*mut JSObject>,
                 object: *mut JSObject)
                 -> Result<Self, ()> {
+        if object.is_null() {
+            return Err(());
+        }
         unsafe {
             let mut guard = RootedGuard::new(cx, root, object);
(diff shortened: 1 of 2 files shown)


src/typedarray.rs
```diff
@@ -70,6 +70,9 @@ impl<'a, T: TypedArrayElement> TypedArray<'a, T> {
                 root: &'a mut Rooted<*mut JSObject>,
                 object: *mut JSObject)
                 -> Result<Self, ()> {
+            if object == ptr::null_mut() {
```
emilio: nit: deindent four spaces, and use `if object.is_null()` instead.

emilio:
Perfect! Please squash the commit (bors doesn't do that automatically), and this is good to merge :)

Issue #336, opened 2017-02-09
Modify TypedArray::create API to accept either a length or an initial slice


jdm:
For cases when we create a new typed array with an initial slice, we always want to use the length of the provided slice. Passing the same length as another argument is redundant; we should use a custom enum argument instead (eg. `CreateWith::Length(5)` and `CreateWith::Slice(foo)`).

Code: `src/typedarray.rs`
Tests: `cargo test --name typedarray` (`tests/typedarray.rs`)

simon-whitehead:
This looks like a great first PR. Mind if I give it a shot? I may need some guidance around code/test style but hopefully I won't need too much hand-holding.

jdm:
Please do!

simon-whitehead:
@jdm Are you able to give me some guidance/documentation around versioning? I imagine I'll be required to bump the crate version?

jdm:
We use a git dependency for rust-mozjs the crate version doesn't actually matter for that case. This crate is also not published to crates.io, so we generally have not been updating the version.

KiChjang:
Fixed by #337.

PR #337, opened 2017-02-11
TypedArray::create API accept length or slice

simon-whitehead:
Issue: https://github.com/servo/rust-mozjs/issues/336

Changes `TypedArray::create` signature to accept a new `CreateWith` argument in place of the `length` and `data` arguments.

Includes similar tests to what was already there using the new enum.

Some notes:

* My first Servo PR. Yay!
* Rustfmt seems to have made some changes. Not sure if thats a problem. The license `/* */` block comments are now `//` single line comments. And some other multi-line code has been changed to single line code. If this is an issue let me know and I'll disable rustfmt and revert those changes back.
* I didn't bump the crate version. I am not sure about the process around that and looking at other PRs I've noticed none of those bumped the version number either.
* I don't yet understand how this applies to Servo. From what I can see everything uses the macro and not `TypedArray::create` directly. I may have missed something though.

Feedback welcome.

@jdm - tagging you since you opened the issue (I hope thats okay?)

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/337)

### DIFF
--- a/src/typedarray.rs
+++ b/src/typedarray.rs
@@ -1,6 +1,6 @@
-/* This Source Code Form is subject to the terms of the Mozilla Public
- * License, v. 2.0. If a copy of the MPL was not distributed with this
- * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 //! High-level, safe bindings for JS typed array APIs. Allows creating new
 //! typed arrays or wrapping existing JS reflectors, and prevents reinterpreting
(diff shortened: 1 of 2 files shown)


simon-whitehead:
Also one other potential noteworthy change.

The original tests allowed different `length` and `data` length inputs. For example, `create` with a length of 10 but a slice with length 5 would result in a 0 padded slice with a length of 10.

The issue itself seems to suggest this shouldn't be possible and so after the change was made, this situation wasn't possible and I have updated the existing tests to reflect that.

jdm:
There are uses of ArrayBuffer::create and Uint8ClampedArray::Create and in Servo right now, and https://github.com/servo/servo/pull/15427 adds more typed array creation.

jdm:
This looks great! Thank you for working on this!
@bors-servo: r+

simon-whitehead:
No problem! Thanks for merging so quickly! :smiley:

(Ps: this was such a painless process. I will look out for more issues I can help with :+1:)

PR #341, opened 2017-03-16
Use Heap for typed arrays.

Ms2ger:


nox:
@jdm Are you satisfied with this patch? It seems ok to me.

nox:
@jdm Ping? I will need this for MSE.

nox:
Actually I think that this should probably be implemented in Servo itself, given there is nothing that is rooted in such a way here in this crate.

PR #344, opened 2017-03-27
Call JS_ShutDown when all runtimes have been dropped.

jdm:
This will enable better investigation of Servo's shutdown behaviour, since currently all of the JS helper threads just sit there until process exit.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/344)

### DIFF
--- a/src/rust.rs
+++ b/src/rust.rs
@@ -16,13 +16,14 @@ use std::default::Default;
 use std::ops::{Deref, DerefMut};
 use std::cell::{Cell, UnsafeCell};
 use std::marker::PhantomData;
+use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
 use consts::{JSCLASS_RESERVED_SLOTS_MASK, JSCLASS_GLOBAL_SLOT_COUNT};
 use consts::{JSCLASS_IS_DOMJSCLASS, JSCLASS_IS_GLOBAL};
 use jsapi;
 use jsapi::{AutoIdVector, AutoObjectVector, CallArgs, CompartmentOptions, ContextFriendFields};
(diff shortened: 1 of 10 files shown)


nox:
> It is currently not possible to initialize SpiderMonkey multiple times (that is, calling JS_Init, then other JSAPI methods, then JS_ShutDown in that order, then doing so again). This restriction may eventually be lifted.

https://developer.mozilla.org/en-US/docs/Mozilla/Projects/SpiderMonkey/JSAPI_reference/JS_ShutDown

jdm:
Yes, this PR is made under the assumption that we do not create new Runtime objects after all existing Runtime objects have been dropped. I could make that explicit by panicking in Runtime::new or making it return a Result if you'd like.

nox:
Yeah, such assumptions are bad in safe code.

jdm:
Fixed.

chrmod:
This PR introduced a breaking change to rust-mozjs api.

Is there any publishing model you use, to help lock projects on certain version?

tschneidereit:
>
> Is there any publishing model you use, to help lock projects on certain
> version?
>
Not yet: mozjs shouldn't at all be considered stable at this point. Once it
gets closer to 1.0, it'll start following semantic versioning.

chrmod:
that is good to know, thanks. Is there anyway to track your progress? How far away you are from 1.0?

jdm:
I would not bet on 1.0 being declared this year.

tschneidereit:
> I would not bet on 1.0 being declared this year.

Agreed, it's quite unlikely it'll get stabilized this year. The API is very likely to undergo substantial changes, with stabilization beginning late this year or in 2018.

PR #353, opened 2017-04-18
added servojsprincipal skeleton

avadacatavra:
Cross origin wrappers (https://github.com/servo/servo/pull/16501) require JSPrincipals. This PR adds the skeleton and basic functionality required.

CC: @jdm

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/353)

### DIFF
--- a/src/glue.rs
+++ b/src/glue.rs
@@ -171,6 +171,7 @@ pub struct ForwardingProxyHandler {
 impl ::std::default::Default for ForwardingProxyHandler {
     fn default() -> ForwardingProxyHandler { unsafe { ::std::mem::zeroed() } }
 }
+
 extern "C" {
     pub fn InvokeGetOwnPropertyDescriptor(handler: *const ::libc::c_void,
                                           cx: *mut JSContext,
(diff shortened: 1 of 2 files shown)


src/jsglue.cpp
```diff
@@ -332,6 +332,30 @@ class WrapperProxyHandler : public js::Wrapper
     }
 };
 
+//http://searchfox.org/mozilla-central/source/js/public/Principals.h#24
```
jdm: I don't think this link adds much.

src/jsglue.cpp
```diff
@@ -332,6 +332,30 @@ class WrapperProxyHandler : public js::Wrapper
...
+class ServoJSPrincipal : public JSPrincipals
+{
+    const void* origin; //box with origin in it
+  public:
+    ServoJSPrincipal(const void* origin)
```
jdm: We will need to accept two function pointers as arguments here - one for `destroy` (that accepts a `*mut JSPrincipals` argument), and one for `write`. This will allow Rust code to provide callbacks that those C++ methods can invoke.

src/jsglue.cpp
```diff
@@ -407,6 +431,16 @@ class ForwardingProxyHandler : public js::BaseProxyHandler
 
 extern "C" {
 
+JSPrincipals *
+CreateServoJSPrincipal(const void* origin){
+  return new ServoJSPrincipal(origin);
+}
+
+const void*
+getPrincipalOrigin(JSPrincipals* principal) {
```
jdm: Let's capitalize the first letter to match.

src/jsglue.cpp
```diff
@@ -407,6 +431,16 @@ class ForwardingProxyHandler : public js::BaseProxyHandler
 
 extern "C" {
 
+JSPrincipals *
```
jdm: nit: remove the space before *

src/glue.rs
```diff
@@ -201,7 +202,14 @@ extern "C" {
      -> *const ::libc::c_void;
     pub fn CreateWrapperProxyHandler(aTraps: *const ProxyTraps)
      -> *const ::libc::c_void;
+    pub fn CreateServoJSPrincipal(origin: *const ::libc::c_void,
+                                  destroy: Option<*const ::libc::c_void>,
+                                  write: Option<bool>)
```
jdm: The destroy and write arguments don't match the C types. See the ones in ProxyTypes for good models.

asajeffrey:
Annoying bikeshedding... is `rust-mozjs` intended to be servo-specific, or general rust bindings to SM for anyone to play with? If the latter, should `ServoJSPrincipal` have a less servo-specific name?

jdm:
RustJSPrincipal would be fine.

src/glue.rs
avadacatavra: i think i fixed this

jdm:
Squash and r=me.

avadacatavra:
r=@jdm

fitzgen:
> is rust-mozjs intended to be servo-specific, or general rust bindings to SM for anyone to play with?

Eventually, it should be for general SM embedding. There is a fair amount of work that needs to happen to make this a reality after the bindings land in m-c, still. A lot of the generic GC-related infrastructure is in Servo's dom component, for example.

PR #377, opened 2017-10-31
Cargo.toml changes needed for publishing on crates.io

### LINKED ISSUE #376: Publish on crates.io
It would be nice to play well with the Rust ecosystem. This is a blocker to publishing anything that depends on mozjs, e.g. https://github.com/asajeffrey/josephine/issues/35

cc: @jdm @tschneidereit @fitzgen @nox

fitzgen:
r? @asajeffrey

Fixes #376 

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/377)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,5 +1,7 @@
 [package]
-name = "js"
+name = "mozjs"
+description = "Rust bindings to the Mozilla SpiderMonkey JavaScript engine."
+repository = "https://github.com/servo/rust-mozjs"
 version = "0.1.7"
 authors = ["The Servo Project Developers"]
 build = "build.rs"
@@ -34,11 +36,9 @@ doctest = false
 debugmozjs = ['mozjs_sys/debugmozjs']
 promises = ['mozjs_sys/promises']
 
-[dependencies.mozjs_sys]
-git = "https://github.com/servo/mozjs"
-
 [dependencies]
 lazy_static = "0.2.1"
 libc = "0.2"
 log = "0.3"
 num-traits = "0.1.32"
+mozjs_sys = "0.50.0"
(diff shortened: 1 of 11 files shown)


Cargo.toml
```diff
@@ -34,11 +36,9 @@ doctest = false
...
 
-[dependencies.mozjs_sys]
-git = "https://github.com/servo/mozjs"
-
 [dependencies]
 lazy_static = "0.2.1"
 libc = "0.2"
 log = "0.3"
 num-traits = "0.1.32"
+mozjs_sys = "0.0.0"
```
KiChjang: The mozjs_sys version on crates.io is 0.1.0 instead of 0.0.0. https://crates.io/crates/mozjs_sys

fitzgen: Huh. I didn't realize it had ever been published before.

fitzgen: @SimonSapin, can you add myself and others (jdm, nox, asajeffrey, KiChjang, etc?) as owners? Thanks!

SimonSapin: @fitzgen I’ve added you. The `servo/cargo-publish` team already has access.

SimonSapin: Since we’re moving things around anyway we could pick one of `mozjs` vs `js` vs something else for both crates?

fitzgen: D'oh -- I meant to rename this crate to `mozjs` since `js` is already taken on crates.io. Good eye!

fitzgen:
I didn't change the version for this crate -- did we want to do that before publishing?

asajeffrey:
I guess we can't keep the version in sync with the SM version, unless we want to limit ourselves to only releasing breaking changes when we change SM. We may as well leave the version alone.

asajeffrey:
Once the crate is published, we can merge https://github.com/servo/servo/pull/19138

PR #381, opened 2017-11-24
Safe, straightforward API for capturing JS stacks

Mrowqa:
PR for https://github.com/servo/servo/issues/14987

Can someone take a look if it goes in the right direction and also look through my comments? Maybe, I have left some "stupid questions", but I wanted to post my changes and get feedback before weekend - I work from Europe, so here is middle of the night.

CC: @jdm

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/381)

### DIFF
--- a/src/rust.rs
+++ b/src/rust.rs
@@ -1259,54 +1259,72 @@ macro_rules! new_jsjitinfo_bitfield_1 {
     }
 }
 
-pub struct CapturedJSStack<'a> { // lifetime inspired by typedarrays
+pub struct CapturedJSStack<'a> {
     cx: *mut JSContext,
     stack: RootedGuard<'a, *mut JSObject>,
 }
 
 impl<'a> CapturedJSStack<'a> {
-    pub fn new(cx: *mut JSContext, max_frame_count: u32) -> Option<Self> { // or maybe another name, for example "capture"?
-        unsafe { // how to handle unsafe in a nice way?
-            rooted!(in(cx) let mut obj = JS_NewPlainObject(cx)); // how to overcome this problem with lifetimes?
-            let mut obj_handle = obj.handle_mut(); // if null, return None // how to handle such errors in a nice way?
-            CaptureCurrentStack(cx, obj_handle, max_frame_count); // if failed, return None
-
-            Some(CapturedJSStack {
-                cx: cx,
-                stack: obj,
-            })
+    pub fn new(cx: *mut JSContext, mut guard: RootedGuard<'a, *mut JSObject>, max_frame_count: u32) -> Option<Self> { // or maybe another name, for example "capture"?
+        let obj_handle = guard.handle_mut(); // if nullptr, return None
+                                             // how to check if handle is not nullptr?
+                                             // how to handle such errors in a nice way?
+        unsafe {
+            if !CaptureCurrentStack(cx, obj_handle, max_frame_count) {
+                return None;
+            }
         }
+
+        Some(CapturedJSStack {
+            cx: cx,
+            stack: guard,
+        })
     }
 
     pub fn as_string(&self, indent: usize) -> Option<String> {
         // TODO where errors can occur? some example of non trivial error handling?
         // TODO how to write tests for this code? Including the problem of obtaining some stack trace
         // do we want to do the actual logic of conversion in this method or maybe in constructor and just return cached result?
-        unsafe {
+        unsafe { // since almost everything here is unsafe, is it okay to wrap it like this?
             let stack_handle = self.stack.handle();
-            let nullptr = 0 as *const ::std::os::raw::c_char;
-            rooted!(in(self.cx) let mut js_string = JS_NewStringCopyZ(self.cx, nullptr)); // can fail, check it
-            let string_handle = js_string.handle_mut();
+            rooted!(in(self.cx) let mut js_string = JS_NewStringCopyZ(self.cx, ptr::null()));
+            let string_handle = js_string.handle_mut(); // as above -- how to check if is not nullptr?
 
-            BuildStackString(self.cx, stack_handle, string_handle, indent); // returns bool, check it
+            if !BuildStackString(self.cx, stack_handle, string_handle, indent) {
+                return None;
+            }
+
             // wanted somehow to just get the utf-16 string and use String::from_utf16_lossy
             // but JS_GetTwoByteStringCharsAndLength requires some object of strange type (AutoCheckCannotGC)
             let str_contents = JS_EncodeStringToUTF8(self.cx, string_handle.handle());
-
-            if str_contents as *const _ == nullptr {
-                None
+            if str_contents.is_null() {
+                return None;
             }
-            else {
-                let mut size = 0 as usize;
-                while *str_contents.offset(size as isize) != 0 { // what is some better way of calculating length on raw pointers?
-                    size += 1;
-                }
 
-                let str_slice = slice::from_raw_parts(str_contents as *const u8, size);
-                let result = String::from_utf8_lossy(str_slice).to_string();
-                JS_free(self.cx, str_contents as *mut ::std::os::raw::c_void);
-                Some(result)
+            let mut size = 0 as usize;
+            while *str_contents.offset(size as isize) != 0 { // what is some better way of calculating length on raw pointers?
+                size += 1;
             }
+
+            let str_slice = slice::from_raw_parts(str_contents as *const u8, size);
+            let result = String::from_utf8_lossy(str_slice).to_string();
+            JS_free(self.cx, str_contents as *mut ::std::os::raw::c_void);
+            Some(result)
         }
     }
 }
+
+#[macro_export]
+macro_rules! capture_stack {
+    ($cx:expr, $max_frame_count:expr) => {
+        rooted!(in($cx) let mut __obj = JS_NewPlainObject($cx));
+        $crate::rust::CapturedJSStack::new($cx, __obj, $max_frame_count)
+    };
+    ($cx:expr) => {
+        capture_stack!($cx, 0)
+    }
+}
+
+// Usage:
+// let stack = capture_stack!(cx);
+// println!(stack.as_string());
--- a/src/rust.rs
+++ b/src/rust.rs
@@ -36,6 +36,7 @@ use jsapi::{RootedBase, RuntimeOptionsRef, SetWarningReporter, Symbol, ToBoolean
 use jsapi::{ToInt32Slow, ToInt64Slow, ToNumberSlow, ToStringSlow, ToUint16Slow};
 use jsapi::{ToUint32Slow, ToUint64Slow, ToWindowProxyIfWindow, UndefinedHandleValue};
 use jsapi::{Value, jsid, PerThreadDataFriendFields, PerThreadDataFriendFields_RuntimeDummy};
+use jsapi::{CaptureCurrentStack, BuildStackString, JS_NewStringCopyZ, JS_EncodeStringToUTF8, JS_free, JS_NewPlainObject};
 use jsval::{ObjectValue, UndefinedValue};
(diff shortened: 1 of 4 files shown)


src/rust.rs
```diff
@@ -1257,3 +1258,55 @@ macro_rules! new_jsjitinfo_bitfield_1 {
...
+        }
+    }
+
+    pub fn as_string(&self, indent: usize) -> Option<String> {
+        // TODO where errors can occur? some example of non trivial error handling?
+        // TODO how to write tests for this code? Including the problem of obtaining some stack trace
+        // do we want to do the actual logic of conversion in this method or maybe in constructor and just return cached result?
+        unsafe {
+            let stack_handle = self.stack.handle();
+            let nullptr = 0 as *const ::std::os::raw::c_char;
```
emilio: nit: rust has `ptr::null()` and `is_null`, so you should use them instead probably.

Mrowqa:
Thank you, @emilio.

I met with @Xanewok and the rest of guys and we came up with the idea of `capture_stack!(cx)` macro. I have also left some other comments waiting for answers.

src/rust.rs
```diff
@@ -1257,3 +1258,73 @@ macro_rules! new_jsjitinfo_bitfield_1 {
...
     }
 }
+
+pub struct CapturedJSStack<'a> {
+    cx: *mut JSContext,
+    stack: RootedGuard<'a, *mut JSObject>,
+}
+
+impl<'a> CapturedJSStack<'a> {
+    pub fn new(cx: *mut JSContext, mut guard: RootedGuard<'a, *mut JSObject>, max_frame_count: u32) -> Option<Self> { // or maybe another name, for example "capture"?
```
jdm: This needs to be `unsafe`, because it accepts a raw pointer argument. I think `new` is a fine name. However, let's make max_frame_count an Option value and use `unwrap_or(0)` to make it easier to use.

src/rust.rs
```diff
@@ -1257,3 +1258,73 @@ macro_rules! new_jsjitinfo_bitfield_1 {
...
+        let obj_handle = guard.handle_mut(); // if nullptr, return None
```
jdm: `if obj_handle.get().is_null()`

src/rust.rs
```diff
@@ -1257,3 +1258,73 @@ macro_rules! new_jsjitinfo_bitfield_1 {
...
+            cx: cx,
+            stack: guard,
+        })
+    }
+
+    pub fn as_string(&self, indent: usize) -> Option<String> {
+        // TODO where errors can occur? some example of non trivial error handling?
+        // TODO how to write tests for this code? Including the problem of obtaining some stack trace
+        // do we want to do the actual logic of conversion in this method or maybe in constructor and just return cached result?
+        unsafe { // since almost everything here is unsafe, is it okay to wrap it like this?
```
jdm: Yes.

src/rust.rs
```diff
@@ -1257,3 +1258,73 @@ macro_rules! new_jsjitinfo_bitfield_1 {
...
+            let stack_handle = self.stack.handle();
+            rooted!(in(self.cx) let mut js_string = JS_NewStringCopyZ(self.cx, ptr::null()));
+            let string_handle = js_string.handle_mut(); // as above -- how to check if is not nullptr?
+
+            if !BuildStackString(self.cx, stack_handle, string_handle, indent) {
+                return None;
+            }
+
+            // wanted somehow to just get the utf-16 string and use String::from_utf16_lossy
+            // but JS_GetTwoByteStringCharsAndLength requires some object of strange type (AutoCheckCannotGC)
```
jdm: https://github.com/servo/rust-mozjs/blob/2b49303d32598a4b00010d6d159ec728ef0145e3/src/conversions.rs#L515-L519

src/rust.rs
```diff
@@ -1257,3 +1258,73 @@ macro_rules! new_jsjitinfo_bitfield_1 {
...
+            let str_contents = JS_EncodeStringToUTF8(self.cx, string_handle.handle());
+            if str_contents.is_null() {
+                return None;
+            }
+
+            let mut size = 0 as usize;
+            while *str_contents.offset(size as isize) != 0 { // what is some better way of calculating length on raw pointers?
```
jdm: For future reference, CStr::from_ptr is useful here.

src/rust.rs
```diff
@@ -1257,3 +1258,73 @@ macro_rules! new_jsjitinfo_bitfield_1 {
...
+        })
+    }
+
+    pub fn as_string(&self, indent: usize) -> Option<String> {
+        // TODO where errors can occur? some example of non trivial error handling?
+        // TODO how to write tests for this code? Including the problem of obtaining some stack trace
+        // do we want to do the actual logic of conversion in this method or maybe in constructor and just return cached result?
+        unsafe { // since almost everything here is unsafe, is it okay to wrap it like this?
+            let stack_handle = self.stack.handle();
+            rooted!(in(self.cx) let mut js_string = JS_NewStringCopyZ(self.cx, ptr::null()));
```
jdm: It should be possible to just use `rooted!(in(self.cx) let mut js_string);`.

src/rust.rs
```diff
@@ -1257,3 +1258,73 @@ macro_rules! new_jsjitinfo_bitfield_1 {
...
+            }
+        }
+
+        Some(CapturedJSStack {
+            cx: cx,
+            stack: guard,
+        })
+    }
+
+    pub fn as_string(&self, indent: usize) -> Option<String> {
```
jdm: Use an Option for indent, and unwrap_or(0).

src/rust.rs
```diff
@@ -1257,3 +1258,73 @@ macro_rules! new_jsjitinfo_bitfield_1 {
...
+            JS_free(self.cx, str_contents as *mut ::std::os::raw::c_void);
+            Some(result)
+        }
+    }
+}
+
+#[macro_export]
+macro_rules! capture_stack {
+    ($cx:expr, $max_frame_count:expr) => {
+        rooted!(in($cx) let mut __obj = JS_NewPlainObject($cx));
```
jdm: We should be able to do `rooted!(in($cx) let mut __obj);` instead.

src/rust.rs
```diff
@@ -1257,3 +1258,73 @@ macro_rules! new_jsjitinfo_bitfield_1 {
...
+        }
+
+        Some(CapturedJSStack {
+            cx: cx,
+            stack: guard,
+        })
+    }
+
+    pub fn as_string(&self, indent: usize) -> Option<String> {
+        // TODO where errors can occur? some example of non trivial error handling?
```
jdm: We should verify that [IsSavedFrame](https://github.com/servo/mozjs/blob/3f38435122538e4b66f88878302f81331c0b6a44/mozjs/js/src/jsapi.h#L6039) returns true.

src/rust.rs
```diff
@@ -1257,3 +1258,73 @@ macro_rules! new_jsjitinfo_bitfield_1 {
...
+        // TODO how to write tests for this code? Including the problem of obtaining some stack trace
```
jdm: We can create an automated test that evaluates some JS that creates several JS stack frames (by defining and invoking functions), then invokes a function that calls into Rust and creates a stack trace. https://github.com/servo/rust-mozjs/blob/master/tests/callback.rs is a good model for this.

src/rust.rs
Mrowqa: Right now, we can't. Please take a look at the definition of the macro: https://github.com/servo/rust-mozjs/blob/master/src/rust.rs#L454-L464.
Should I expand it and if yes, would you prefer to do it in this PR or separate one? Another problem is that `JS_NewPlainObject` returns `*mut JSObject` and `JS_NewCopyStringZ`, which I used somewhere else, returns `*mut JSString`. So, what should be the type of the new binding created within the macro?

src/rust.rs
Mrowqa: Are you sure about passing null pointer to the `JS_GetTwoByteStringCharsAndLength`? In C++ it requires a reference: https://developer.mozilla.org/en-US/docs/Mozilla/Projects/SpiderMonkey/JSAPI_Reference/JS_GetLatin1StringCharsAndLength. Maybe it doesn't have any internal fields, but isn't it undefined behavior?

I read the comment before definition (https://dxr.mozilla.org/mozilla-central/source/js/public/GCAPI.h?q=AutoCheckCannotGC&redirect_type=direct#852-876), but I am still not sure when we can use it. Does it mean that we can use it when we have rooted all the objects whose handles we are passing to the function?

src/rust.rs
jdm: Oh, my mistake. It should be possible to use a null pointer instead, though.

src/rust.rs
Mrowqa: Isn't it guaranted by [CaptureCurrentStack](https://dxr.mozilla.org/mozilla-central/source/js/src/jsapi.cpp?q=CaptureCurrentStack&redirect_type=direct#7700) after returning `true`?

jdm: Yes, but if a consumer does not use the macros and overwrites the value in the `Rooted<*mut JSObject>` before calling as_string, it won't work. This is probably an instance where we would get a `false` value from the code in as_string.

src/rust.rs
Mrowqa: So, if some function expects a mutable handle for an output argument, we can just pass a handle to a null pointer? Did I get it right?

src/rust.rs
KiChjang: If you look at where the `nogc` parameter is actually used in the function body, you'll see that it's passed as a parameter to [twoByteChars](https://dxr.mozilla.org/mozilla-central/source/js/src/vm/String.h#681), and that function doesn't do anything with it, so I'm quite confident that there's no UB going on here.

In fact, the comment before the definition of AutoCheckCannotGC states that this is only used for static rooting hazard analysis, so I don't see how it can go wrong when you send in a null pointer.

Mrowqa: @KiChjang thank you for the explanation. I wasn't sure.

src/rust.rs
jdm: Yes. The mutable handle allows the function to initialize it with a meaningful value.

Mrowqa:
Now it should be almost complete. When I use `JS_GetTwoByteStringCharsAndLength` in `as_string()`, the method returns some trash instead of actual stack trace. Probably some problem with text encoding. I left this code commented out. Version with `JS_EncodeStringToUTF8` works.

Mrowqa:
@jdm I have just removed the commented out code with `JS_GetTwoByteStringCharsAndLength`. To be honest, I am really confused what it does. I thought JS Engine should keep internally strings encoded in UTF-16 and this function looked like getting a pointer to the internal buffer since it doesn't allocate anything and return `const char16_t*`. I have printed out the memory and it contains the expected stack, but it looks like it is encoded in UTF-8.

So, since I am really confused what is going on, I have left the version with [JS_EncodeStringToUTF8](https://developer.mozilla.org/en-US/docs/Mozilla/Projects/SpiderMonkey/JSAPI_Reference/JS_EncodeString) since it works as expected. The drawback of this is that we have this one extra `JS_Free` (and associated memory allocation). Moreover, it just cuts out the higher byte of character instead of converting it to, for instance, `'?'`.

We can merge this, if it is okay with you. And I am willing to learn more about what is going on internally, what these functions are supposed to do and so on :)

Mrowqa:
@jdm Travis failed to have installed nightly toolchain, but the code actually works and passes the tests.

jdm:
With respect to the unexpected output of JS_GetTwoByteStringCharsAndLength, that sounds like what the code at https://github.com/Mrowqa/rust-mozjs/blob/e654bb6c238b659e85ac087adf6568eda20e8648/src/conversions.rs#L510-L513 is intended to address. Perhaps we can try extracting https://github.com/Mrowqa/rust-mozjs/blob/e654bb6c238b659e85ac087adf6568eda20e8648/src/conversions.rs#L511-L519 into a method that we can reuse here?

PR #382, opened 2017-11-27
Reimplement CustomAutoRooter hierarchy

Xanewok:
This is an attempt at https://github.com/servo/servo/issues/16678. It doesn't yet contain conversion code that's described here: https://github.com/servo/servo/issues/16678#issuecomment-345774906.

I wanted to see if the design is somewhat sound and ergonomic. Tried to make CustomAutoRooter a convenient generic type that could support tracing arbitrary types, just like C++ equivalent does, albeit with generics, since we can't just use C++ inheritance.

Here `SequenceRooter<T>` is implemented as a simple [alias](https://github.com/Xanewok/rust-mozjs/blob/custom-auto-rooter/src/rust.rs#L592) for `CustomAutoRooter<Vec<T>>`.

cc @jdm

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/382)


src/rust.rs
```diff
@@ -463,6 +464,185 @@ macro_rules! rooted {
...
+unsafe impl CustomTrace for Vec<*mut JSObject> {
+    fn trace(&self, trc: *mut JSTracer) {
+        for obj in self {
+            let obj = obj as *const *mut _ as *mut *mut _;
+            unsafe { CallObjectRootTracer(trc, obj, c_str!("sequence<object>")); }
+        }
+    }
+}
+
+unsafe impl CustomTrace for  Vec<Value> {
```
jdm: nit: extra space after `for`

src/rust.rs
```diff
@@ -463,6 +464,185 @@ macro_rules! rooted {
...
+
+unsafe impl<T: CustomTrace> CustomTrace for Vec<Option<T>> {
+    fn trace(&self, trc: *mut JSTracer) {
+        for seq in self.iter().filter_map(|x| x.as_ref()) {
+            seq.trace(trc);
+        }
+    }
+}
+
+unsafe impl<T: CustomTrace> CustomTrace for Vec<T> {
```
jdm: I think it makes more sense to implement CustomTrace for the following types:
* `*mut Object`
* `Value`
* `Option<T>` where `T: CustomTrace`
* `Vec<T>` where `T: CustomTrace`

This composes better than focusing on different specializations of `Vec<T>`.

src/rust.rs
```diff
@@ -463,6 +464,185 @@ macro_rules! rooted {
...
+        }
+    }
+}
+
+// TODO: Implement other CustomTraces
+// See: https://dxr.mozilla.org/mozilla-central/rev/ee21e5f7f1c1726e0ed2697eb45df54cdceedd36/dom/bindings/BindingUtils.h#2289
+
+pub type SequenceRooter<T> = CustomAutoRooter<Vec<T>>;
+
+/// TODO: Document me (similar to Rooted)
```
jdm: Yes please!

src/rust.rs
```diff
@@ -463,6 +464,185 @@ macro_rules! rooted {
...
+pub struct SequenceRooterGuard<'a, T: 'a + RootKind + GCMethods> {
```
jdm: For some of the work that the rest of your group is doing, we will want to make this more general. I'm fine if you want to make this follow-up work.

Since there is nothing special about sequences here, let's make this more general: `RooterGuard<T>` where `T: StackRoot` (which can be a new type that provides add_to_root_stack and remove_from_stack, as well as a way to access the contained data).

tests/sequence_rooter.rs
```diff
@@ -0,0 +1,74 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#![cfg(feature = "debugmozjs")]
```
jdm: Let's remove this so the test is always run.

tests/sequence_rooter.rs
```diff
@@ -0,0 +1,74 @@
...
+/// by checking if appropriate virtual trace function was called.
+#[test]
+fn custom_auto_rooter_vftable() {
+    static mut TRACE_FN_WAS_CALLED: bool = false;
+
+    pub struct TestStruct { }
+
+    unsafe impl CustomTrace for TestStruct {
+        fn trace(&self, _: *mut JSTracer) {
+            unsafe { TRACE_FN_WAS_CALLED = true; }
```
jdm: We could store a `Cell<bool>` inside CustomTrace instead of using `static mut` for this.

tests/sequence_rooter.rs
```diff
@@ -0,0 +1,74 @@
...
+/// C++ virtual trace function is called on Rust side.
+#[test]
+fn sequence_rooter_vftable() {
+    static mut TRACE_FN_WAS_CALLED: bool = false;
+
+    pub struct TestStruct { }
+
+    unsafe impl CustomTrace for TestStruct {
+        fn trace(&self, _: *mut JSTracer) {
+            unsafe { TRACE_FN_WAS_CALLED = true; }
```
jdm: Same as previous comment.

tests/sequence_rooter.rs
```diff
@@ -0,0 +1,74 @@
...
+
+    let rt = Runtime_::new().unwrap();
+
+    unsafe {
+        let mut rooted = mozjs::rust::CustomAutoRooter::<TestStruct>::new();
+        rooted.data = Some(TestStruct { });
+
+        rooted.add_to_root_stack(rt.cx());
+        JS_GC(rt.rt());
+        rooted.remove_from_root_stack();
```
jdm: This test would benefit from the generalization I suggested earlier.

src/rust.rs
```diff
@@ -463,6 +464,185 @@ macro_rules! rooted {
...
+    pub fn new_unrooted(tag: AutoGCRooterTag) -> AutoGCRooter {
+        AutoGCRooter {
+            down: ptr::null_mut(),
+            tag_: tag as isize,
+            stackTop: ptr::null_mut(),
+        }
+    }
+
+    pub unsafe fn add_to_root_stack(&mut self, cx: *mut JSContext) {
+        let autoGCRooters: &mut _ = {
```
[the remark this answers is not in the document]
Xanewok: Yes, because we need a pointer to the `autoGCRooters_` stack list itself (which is also represented as a pointer).
The `stackTop` member is initialized with it in C++: https://dxr.mozilla.org/mozilla-central/source/js/public/RootingAPI.h#839.

src/rust.rs
Xanewok: I used specialization to retain the `"sequence<any>"` and `"sequence<object>"` strings, as used in C++: https://dxr.mozilla.org/mozilla-central/source/dom/bindings/BindingUtils.h#2163.

Should I just provide `"object"` and `"any"` messages for `*mut JSObject` and `Value` and use that in `Vec<T>` where `T: CustomTrace` instead?

tests/sequence_rooter.rs
Xanewok: Ah, `JS_SetGCZeal` needed that and I thought I stumbled upon this also with `JS_GC`, but that's not the case and it's not needed here. Will do, thanks!

tests/sequence_rooter.rs
Xanewok: As we talked, `#[cfg(test)]` code is not compiled in a library target that's used by a Cargo integration test, so we can't modify the `CustomTrace` trait here.

I can extract the boilerplate code to a macro that declares the static variable and the impl block, if that works.

Xanewok:
Addressed feedback and pushed related changes.

tests/sequence_rooter.rs
jdm: Whoops, I meant to says `TestStruct` instead of `CustomTrace`.

src/rust.rs
jdm: Yeah, I don't think it's important to get the sequence strings here.

Xanewok:
@jdm pushed a commit that uses `Cell` instead of `static mut` variable. Could you take another look at it and tell me if this needs more work still?

src/rust.rs
```diff
@@ -463,6 +465,196 @@ macro_rules! rooted {
...
+pub type SequenceRooterGuard<'a, T> = CustomAutoRooterGuard<'a, Vec<T>>;
+
+#[macro_export]
+macro_rules! rooted_seq {
+    (in($cx:expr) let $name:ident = $init:expr) => {
+        let mut __root = $crate::rust::CustomAutoRooter::new($init);
+        let $name = __root.root($cx);
+    };
+    (in($cx:expr) let mut $name:ident = $init:expr) => {
+        let mut __root = $crate::rust::SequenceRooter::new($init);
```
jdm: One of these is CustomAutoRooter, and one is SequenceRooter. That's surprising!

Xanewok: Oops! Fixed that now, renamed the macro to a more general `auto_root` name, to be used with CustomAutoRooter itself.

jdm:
@bors-servo: r+
Thanks!

jdm:
Ah, having both new tests in the same file is hitting the usual error.
```
thread 'root_macro' panicked at 'called `Result::unwrap()` on an `Err` value: ()', src\libcore\result.rs:906:4
```

Xanewok:
Darn it, I tried to reproduce it and ran `cargo test` ~2000 times on Ubuntu 16.10 on i7-3770K, but couldn't reproduce it in the end.
@jdm do you think we can proceed with this and work in the meantime on bulletproofing the testing infra? Maybe we can do a test run with `cargo test -- --test-threads=1` on CI to see if that helps?

jdm:
If you move the second new test into a separate file, it should avoid the problem entirely.

Xanewok:
Done. However I'm not sure if that's a good idea moving forward, since linking and producing separate binary for each test takes considerable amount of time, so if each test doesn't need a completely separate process, it'd probably be faster and easier to maintain to run tests using only a single thread instead in fewer test binaries.

Xanewok:
Same as https://github.com/servo/rust-mozjs/pull/381#issuecomment-348116907, Travis couldn't start macOS debugmozjs job for a long time and when it did, it couldn't set up the environment:
```
error: toolchain 'nightly' is not installed
info: caused by: not a directory: '/Users/travis/.rustup/toolchains/nightly-x86_64-apple-darwin'
```
https://travis-ci.org/servo/rust-mozjs/jobs/309158923

PR #383, opened 2017-11-27
Don't rerun build script unnecessarily

### LINKED ISSUE #380: Make build script use rerun-if-changed
Right now any time you modify one of the test files, the whole library needs to be rebuilt. We should scope the rebuild to any file in src/ instead.

Xanewok:
Closes #380. From what I can see the only thing [impacting](https://github.com/servo/rust-mozjs/blob/master/CMakeLists.txt#L7) the cmake build is the jsglue.cpp file, so I set to only watch that file, instead.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/383)

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -8,4 +8,5 @@ fn main() {
     let dst = cmake::Config::new(".").build();
     println!("cargo:rustc-link-search=native={}/lib", dst.display());
     println!("cargo:rustc-link-lib=static=jsglue");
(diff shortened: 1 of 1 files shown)


jdm:
@bors-srrvo r+
Thanks!

jdm:
@bors-servo r+
Thanks!

PR #390, opened 2018-01-21
Add a conversion from the newly-#[stable] `ptr::NonNull` type

SimonSapin:
This will allow removing the `NonNullJSObjectPtr` hack in Servo, but raises the minimum Rust version to 1.25 (currently Nigthly)

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/390)

### DIFF
--- a/src/conversions.rs
+++ b/src/conversions.rs
@@ -38,7 +38,7 @@ use jsapi::{JSContext, JSObject, JSString, MutableHandleValue, RootedObject};
 use jsval::{BooleanValue, Int32Value, NullValue, UInt32Value, UndefinedValue};
 use jsval::{JSVal, ObjectValue, ObjectOrNullValue, StringValue};
 use rust::{ToBoolean, ToInt32, ToInt64, ToNumber, ToUint16, ToUint32, ToUint64};
-use rust::{ToString, maybe_wrap_object_or_null_value};
+use rust::{ToString, maybe_wrap_object_or_null_value, maybe_wrap_object_value};
 use rust::maybe_wrap_value;
 use libc;
 use num_traits::{Bounded, Zero};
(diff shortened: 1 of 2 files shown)


SimonSapin:
@fitzgen @jdm @nox How do you feel about this repository requiring Rust 1.15? According to https://wiki.mozilla.org/Rust_Update_Policy_for_Firefox Gecko can use that version from April 12, but as far as I understand it doesn’t use this repo anyway.

nox:
I assume you meant 1.25? I don't mind.

SimonSapin:
Oops, yes. I mostly ask in case it interferes with plans to unify this repo with `js/` in mozilla-central.

SimonSapin:
In particular: this is only clean-up, so if it requires adding a Cargo feature somewhere it’s not worth it and we can just wait.

src/conversions.rs
```diff
@@ -646,6 +646,15 @@ impl ToJSValConvertible for *mut JSObject {
     }
 }
 
+// https://heycam.github.io/webidl/#es-object
+impl ToJSValConvertible for ptr::NonNull<JSObject> {
+    #[inline]
+    unsafe fn to_jsval(&self, cx: *mut JSContext, rval: MutableHandleValue) {
+        rval.set(ObjectOrNullValue(self.get()));
```
emilio: This can probably use `ObjectValue` and `maybe_wrap_object_value`.

SimonSapin: Done. Indeed, that’s what the equivalent impl in Servo does.

SimonSapin:
I take it "approved" means:

@bors-servo r=jdm

SimonSapin:
Eh, OS X job on Travis-CI still not scheduled after 13 hours…

PR #393, opened 2018-03-04
Add lifetimes for Handle and MutableHandle

### LINKED ISSUE #153: Make Handle objects take lifetimes
Handles are currently dangerous right now since they can outlive the Roots that they're taken from. We should make them take lifetimes to avoid that issue. This api is introduced in https://github.com/servo/rust-mozjs/pull/150

marmistrz:
Servo compiled fine for me with the following patch set. 

Since the `From` trait consumes the object, I created a `raw` method for `MutableHandle`. This is not ideal, since `into()` would be more idiomatic. Is there anything better I could do or should I just leave it as it is?

There are two questions marked with `// REVIEW` since I'm not sure if the explicit lifetimes are the same what compiler inferred.

Closes #153.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/393)


jdm:
This will require some updates to the unit tests, too.

src/conversions.rs
```diff
@@ -180,7 +181,7 @@ fn clamp_to<D>(d: f64) -> D
 // https://heycam.github.io/webidl/#es-void
 impl ToJSValConvertible for () {
     #[inline]
-    unsafe fn to_jsval(&self, _cx: *mut JSContext, rval: MutableHandleValue) {
+    unsafe fn to_jsval(&self, _cx: *mut JSContext, mut rval: MutableHandleValue) {
```
jdm: We should modify the ToJSValConvertible trait to accept a lifetime argument and use it for the rval argument.

src/rust.rs
```diff
@@ -412,10 +424,10 @@ impl<'a, T: 'a + RootKind + GCMethods> RootedGuard<'a, T> {
         }
     }
 
-    pub fn handle(&self) -> Handle<T> {
-        unsafe {
-            Handle::from_marked_location(&self.root.ptr)
-        }
+    // REVIEW: is this equivalent to?
+    // pub fn handle(&self) -> Handle<T>
+    pub fn handle(&'a self) -> Handle<'a, T> {
```
jdm: I believe it is equivalent. Let's use the explicit form.

src/rust.rs
```diff
@@ -707,34 +768,104 @@ impl<T> MutableHandle<T> {
     }
 }
 
-impl<T> Deref for MutableHandle<T> {
+impl<'a, T> MutableHandle<'a, T> {
+    pub unsafe fn from_marked_location(ptr: *mut T) -> Self {
+        RawMutableHandle::from_marked_location(ptr).into()
+    }
+
+    pub fn handle(&self) -> RawHandle<T> {
```
jdm: We should probably return a non-raw handle from this API.

src/rust.rs
```diff
@@ -853,9 +984,9 @@ impl<T: GCMethods + Copy> Heap<T> {
         self.ptr.get()
     }
 
-    pub fn handle(&self) -> Handle<T> {
+    pub fn handle(&self) -> RawHandle<T> {
```
jdm: We should definitely return a non-raw Handle here, and return a MutableHandle from handle_mut that is not using the 'static lifetime.

src/rust.rs
```diff
@@ -707,34 +768,104 @@ impl<T> MutableHandle<T> {
...
+    pub fn undefined() -> &'static Self {
         unsafe {
-            UndefinedHandleValue
+            &UndefinedHandleValue
         }
     }
 }
 
+impl<'a> HandleValue<'a> {
+    pub fn null() -> Self {
```
jdm: Let's be explicit and use `HandleValue<'static'>` for these methods.

src/rust.rs
```diff
@@ -707,34 +768,104 @@ impl<T> MutableHandle<T> {
...
+    }
+}
+
+impl<'a, T> From<MutableHandle<'a, T>> for RawMutableHandle<T> {
+    fn from(handle: MutableHandle<'a, T>) -> Self {
+        unsafe { RawMutableHandle::from_marked_location(handle.ptr) }
+    }
+}
+
+impl<'a, T> From<RawMutableHandle<T>> for MutableHandle<'a, T> {
```
jdm: How important are the `From<Raw>` impls for the safe handles? It makes me uncomfortable that we can create safe handles with 'static lifetimes from arbitrary handles that do not actually have static lifetimes.

src/rust.rs
```diff
@@ -657,40 +668,90 @@ macro_rules! auto_root {
...
+        where T: Copy
+    {
+        *self.ptr
+    }
+
+    pub fn new(ptr: &'a T) -> Self {
+        Handle { ptr: ptr }
+    }
+
+    pub unsafe fn from_marked_location(ptr: *const T) -> Self {
```
jdm: Is this used outside of this crate? If we could avoid making it public, that would be useful.

marmistrz:
Honestly, I don't really have an idea what we could test here. That's just mimicking the Handle's behavior, all the checks are done by the compiler.

jdm:
We could write some [compile_fail doctests](https://github.com/servo/servo/blob/5213e2a5a9ed8859a004e480f52c7e04f8d7f85e/tests/unit/script_plugins/lib.rs#L22) that demonstrate that handle values with lifetimes can't outlive the objects from which they were obtained.

jdm:
Oh, and to be clear when I said "updates to the unit tests", I was specifically referring to the fact that they do not compile any longer according to TravisCI :)

src/rust.rs
marmistrz: removed.

jdm:
The changes here look good to me as long as we can make Servo build with them before merging them.

src/rust.rs
marmistrz: Are you sure that it's not using the `'static` lifetime now? I'm not so sure of it.

jdm: I'm not even sure which code is being referred to any more. I believe that any code with a &self argument that returns a Handle will have the lifetime inferred to match the self argument, but it would be valuable to write that out explicitly.

src/generate_jsapi_wrappers.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh
+# This is one big heuristic but seems to work well enough
```
jdm: This is a clever solution!

src/rust.rs
```diff
@@ -1635,3 +1635,101 @@ macro_rules! capture_stack {
...
+        wrap!(@inner $saved <> ($($acc,)* $arg.into(),) <> $($rest)*);
+    };
+    (@inner $saved:tt <> ($($acc:expr,)*) <> $arg:ident: MutableHandleValue, $($rest:tt)*) => {
+        wrap!(@inner $saved <> ($($acc,)* $arg.into(),) <> $($rest)*);
+    };
+    (@inner $saved:tt <> ($($acc:expr,)*) <> $arg:ident: $type:ty, $($rest:tt)*) => {
+        wrap!(@inner $saved <> ($($acc,)* $arg,) <> $($rest)*);
+    };
+    (@inner ($func_name:ident ($($args:tt)*) -> $outtype:ty) <> ($($argexprs:expr,)*) <> ) => {
+        pub unsafe fn $func_name($($args)*) -> $outtype {
```
jdm: Let's add `#[inline]` here.

src/rust.rs
```diff
@@ -1635,3 +1635,101 @@ macro_rules! capture_stack {
...
+    use jsapi::Shape;
+    use jsapi::SourceBufferHolder;
+    use jsapi::Symbol;
+    use jsapi::SymbolCode;
+    use jsapi::TwoByteChars;
+    use jsapi::Value;
+    use libc::FILE;
+    use super::{Handle, HandleId, HandleObject, HandleValue};
+    use super::{MutableHandle, MutableHandleObject, MutableHandleValue};
+    include!("jsapi_wrappers.rs");
```
jdm: The file in this PR is called jsapi_wrappers.in. Is something missing?

src/conversions.rs
```diff
@@ -238,7 +239,7 @@ unsafe fn convert_int_from_jsval<T, M>(cx: *mut JSContext, value: HandleValue,
 // https://heycam.github.io/webidl/#es-boolean
 impl ToJSValConvertible for bool {
     #[inline]
-    unsafe fn to_jsval(&self, _cx: *mut JSContext, rval: MutableHandleValue) {
+    unsafe fn to_jsval(&self, _cx: *mut JSContext, mut  rval: MutableHandleValue) {
```
nox: Nit: double space.

src/rust.rs
```diff
@@ -1635,3 +1635,101 @@ macro_rules! capture_stack {
...
+        }
+    };
+    (pub fn $func_name:ident($($args:tt)*) -> $outtype:ty) => {
+        wrap!(@inner ($func_name ($($args)*) -> $outtype) <> () <> $($args)* ,);
+    };
+    (pub fn $func_name:ident($($args:tt)*)) => {
+        wrap!(pub fn $func_name($($args)*) -> ());
+    }
+}
+pub mod wrappers {
```
jdm: Let's add documentation like:
```
Wrappers for JSAPI methods that accept FFI Handle and MutableHandle arguments. The wrapped methods are identical except that they accept Handle and MutableHandle arguments that include lifetimes instead.
```

src/rust.rs
marmistrz: that was fixed while merging upstream :)

src/rust.rs
marmistrz: You mean:
> Wrappers for JSAPI methods that accept lifetimed Handle and MutableHandle arguments.
?

nox:
@marmistrz AFAICT skimming through that macro, the only way to require less clauses is to unconditionally transform all arguments, as opposed to just the ones with `MutableHandle` etc. This would require a trait implemented for all arguments found in invocations of `wrap!`. I think this is good enough for now.

src/rust.rs
jdm: My comment scrolls horizontally, so keep reading.

jdm:
Hmm, Unbox seems to be a platform-specific function for some reason:
```
error[E0425]: cannot find function `Unbox` in module `jsapi`
    --> src\rust.rs:1680:23
     |
1680 |         wrap!(@inner ($func_name ($($args)*) -> $outtype) <> () <> $($args)* ,);
     |                       ^^^^^^^^^^ not found in `jsapi`
     | 
    ::: src\jsapi_wrappers.in:3:1
     |
3    | wrap!(pub fn Unbox(cx: *mut JSContext, obj: HandleObject, vp: MutableHandleValue) -> bool);
     | ------------------------------------------------------------------------------------------- in this macro invocation
help: possible candidate is found in another module, you can import it into scope
     |
1693 |     use rust::wrappers::Unbox;
     |

```

Issue #343, opened 2017-03-24
Heap constructor is unsafe


jdm:
The Heap constructor calls `Heap::set` which triggers a post barrier. This stores the current address of the value stored inside the Heap, which will be moved once the Heap constructor returns. I have not yet been able to think of a way to prevent people from calling `Heap::set` before the `Heap` value has stopped being moved, short of making the `Heap` constructor return a Box instead.

MortimerGoro:
I think that I got a new instance of this issue. The gamepad implementation worked great until I moved the typedarray from a setter to the heap constructor after a suggestion in  the PR code review. Now there is a GC related crash in about 30 seconds after running the app.

Here you can see the commit with the failing and working version: https://github.com/MortimerGoro/servo/commit/644114df47e46372990c2d6e28911b3b8543029e

Do you think that the crash is related to this issue? I want to understand why the crash happens.

jdm:
Yes, that's the exact same symptom and solution as https://github.com/servo/servo/issues/15445.

asajeffrey:
Rather drastic suggestion: can we remove the constructor? Or mark it as `unsafe`? The fix in (e.g.) https://github.com/servo/servo/pull/16500/files#diff-ee0d03fc9e0cdadcde40c50bd0f8369eL38 is to replace the constructor by a use of `Heap::default()` followed by an explicit `set` after the heap object has finished moving.

jdm:
Yes, we should remove the constructor.

jdm:
I want to apply this patch:
```diff
diff --git a/src/conversions.rs b/src/conversions.rs
index 2b2f6a9..bd114ce 100644
--- a/src/conversions.rs
+++ b/src/conversions.rs
@@ -216,7 +216,9 @@ impl FromJSValConvertible for Heap<JSVal> {
                          value: HandleValue,
                          _option: ())
                          -> Result<ConversionResult<Self>, ()> {
-        Ok(ConversionResult::Success(Heap::new(value.get())))
+        // FIXME: This is horribly unsound and has no guarantee of working.
+        // https://github.com/servo/rust-mozjs/issues/334
+        Ok(ConversionResult::Success(Heap::do_not_use_this(value.get())))
     }
 }

diff --git a/src/rust.rs b/src/rust.rs
index 8faab9b..8f495f2 100644
--- a/src/rust.rs
+++ b/src/rust.rs
@@ -642,7 +642,7 @@ impl GCMethods for Value {
 }

[...]

PR #395, opened 2018-03-10
Add safe Heap::boxed constructor

Xanewok:
While this doesn't remove old Heap constructor (#343), it adds a new `boxed` function along with some additional explanation while this may be preferred until a better/different solution comes up to construct a Heap value directly with an appropriate value set, in a safe manner.

This made my work (which I'll upload later) on typed arrays, and also indirectly for objects, in webidl unions safer and more convenient, so I think it's a good addition.

r? @jdm

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/395)

### DIFF
--- a/src/rust.rs
+++ b/src/rust.rs
@@ -836,6 +836,21 @@ impl<T: GCMethods + Copy> Heap<T> {
         ptr
     }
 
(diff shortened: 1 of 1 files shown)


jdm:
I agree. This is a useful API to provide.
@bors-servo r+

PR #397, opened 2018-03-13
Add TypedArray API using Heap-wrapped objects

Xanewok:
Needed for https://github.com/servo/servo/pull/20267.

This changes TypedArray wrapper type to only contain typed array-related logic and not handle rooting, and introduces 2 different possible storage options for wrapped objects
- `*mut JSObject`, can be rooted on stack (e.g. using CustomAutoRooter)
-  `Box<Heap<*mut JSObject>>`, can be rooted on heap with `JSTraceable` trait (implemented on Servo side)

Since the API changes are breaking, I bumped minor version so that Servo will not break when updating.

r? @jdm

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/397)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,7 +2,7 @@
 name = "mozjs"
 description = "Rust bindings to the Mozilla SpiderMonkey JavaScript engine."
 repository = "https://github.com/servo/rust-mozjs"
-version = "0.1.14"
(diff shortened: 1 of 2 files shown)


jdm:
@bors-servo r+
Thanks for working on this!

PR #404, opened 2018-03-23
Sanitize Heap::handle(_mut) functions

### LINKED ISSUE #351: Heap::mut_handle avoids the GC barriers
This looks totally unsafe to me. It probably needs to be removed.

Xanewok:
Fixes #351.

Also marked `Heap::handle` as `unsafe`, hopefully the explanation in the code is concise/good enough.

r? @jdm

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/404)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,7 +2,7 @@
 name = "mozjs"
 description = "Rust bindings to the Mozilla SpiderMonkey JavaScript engine."
 repository = "https://github.com/servo/rust-mozjs"
-version = "0.3.2"
+version = "0.4.0"
 authors = ["The Servo Project Developers"]
 build = "build.rs"
 license = "MPL-2.0"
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -2,7 +2,7 @@
 name = "mozjs"
 description = "Rust bindings to the Mozilla SpiderMonkey JavaScript engine."
 repository = "https://github.com/servo/rust-mozjs"
-version = "0.3.2"
+version = "0.3.3"
```
Eijebong: This is a breaking change, right ? Should probably be 0.4

Xanewok: Yeah, I admit I wanted to cheat a little here :( Wanted to avoid the constant churn with bumping the minor versions, but that should be 0.4, you're right!

Eijebong: :eyes:

jdm:
@bors-servo r+
Hooray for getting rid of footguns!

PR #407, opened 2018-03-24
Added Default implementation for PropertyDescriptor

taki-jaro:
It is needed for servo to be able to use `JS_GetOwnPropertyDescriptorById` method as it needs `MutableHandle<PropertyDescriptor>` as output parameter. To get the `MutableHandle` one should do sth like `rooted!(in(cx) let mut desc = dummyPropertyDescriptor)` and then `desc.handle_mut()`. Now there are no easy ways to get `dummyPropertyDescriptor`.

I am not sure if it is the proper way. AFAIK the jsapi_* files are automatically generated they shouldn't be modified and I don't know how to make bindgen generate `derive(Default)` for PropertyDescriptor

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/407)

### DIFF
--- a/src/property_descriptor.rs
+++ b/src/property_descriptor.rs
@@ -1,0 +1,9 @@
+use jsapi::PropertyDescriptor
+use jsapi::UndefinedValue
+use std::ptr::null
+
+impl Default for PropertyDescriptor {
+    fn defuault() -> PropertyDescriptor {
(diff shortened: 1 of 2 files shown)


jdm:
I would just add this to rust.rs. Right now this code won't be compiled unless lib.rs is modified to include the new module.

taki-jaro:
Done. So stupid of me, I forgot to update lib.rs

PR #408, opened 2018-03-24
typedarray: Use a Cell for handling interior mutability.

emilio:
The typed array wrapper has a "computed" option that uses to cache its data.

That means that you need a mutable reference to the typed array in order to read
it, which is not great.

In particular, it's plain impossible when the typed array is provided as a
member of a struct via WebIDL.

Use a `Cell` and expose to_vec and as_slice in non-mutable references.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/408)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,7 +2,7 @@
 name = "mozjs"
 description = "Rust bindings to the Mozilla SpiderMonkey JavaScript engine."
 repository = "https://github.com/servo/rust-mozjs"
-version = "0.4.0"
+version = "0.4.1"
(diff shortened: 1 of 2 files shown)


emilio:
r? @jdm 

 * This is blocking the Bluetooth bindings from being ported to use typed arrays, since the filters come from a borrowed WebIDL struct: https://github.com/servo/servo/blob/0c929ad3fd902f4314844787868fe891033d3b01/components/script/dom/bluetooth.rs#L487

PR #410, opened 2018-03-26
rooted macro now uses default initializer

### LINKED ISSUE #405: Allow rooted! macro to use default initializer when none is specified
https://github.com/servo/rust-mozjs/blob/8fe9be3fabad5ebe95dca9b3eceb4a3a3ac052da/src/rust.rs#L458-L468 currently requires an initializer value. We should make it so that if none is provided we pass Default::default() instead.

christianpoveda:
This is supposed to fix #405

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/410)


KiChjang:
Looks good to me, thanks!

@bors-servo r+

PR #413, opened 2018-03-26
Adds getting started docs for macOS

anthgur:
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/413)

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,3 +3,31 @@
 Rust bindings to SpiderMonkey
 
 [Documentation](http://doc.servo.org/mozjs/)
+
+## Setting up your environment
+
+### Rust
+
+This project requires Rust nightly-2018-03-25 or greater.
+For installation of the Rust toolchain see [`https://rustup.rs/`](https://rustup.rs/).
+
+### Other dependencies
+
+#### macOS
+#### On macOS (Homebrew)
+
+```sh
+brew install automake pkg-config python2 cmake
+```
+
+## Building the project
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -3,3 +3,55 @@
...
+
+```sh
+rustup toolchain install nightly
+rustup default nightly
+```
+
+### Other dependencies
+
+#### OS X
+#### On OS X (homebrew)
```
atouchet: Apple no longer uses the OS X branding. Can these be changed to say macOS instead? Also I think homebrew should be capitalized here.

anthgur: Good catch, thank you!

README.md
```diff
@@ -3,3 +3,55 @@
 Rust bindings to SpiderMonkey
 
 [Documentation](http://doc.servo.org/mozjs/)
+
+## Setting up your environment
+
+### Rust
+
+This project requires Rust nightly-2018-03-25 or greater. You can install this with Rustup.rs:
```
jdm: I don't think it makes sense to repeat the rustup installation instructions here. It's more efficient to direct people to https://rustup.rs/.

anthgur: That makes sense, can we keep the nightly info though? I spent some time figuring out I needed to use nightly instead of stable.

jdm: Yes, that's fine.

jdm:
@bors-servo retry
* #416

PR #423, opened 2018-04-11
Require C++14

vyv03354:
Currently C++14 features are unavailable due to explicit -std=c++11 switch.
See also https://bugzilla.mozilla.org/show_bug.cgi?id=1451931

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/423)


KiChjang:
Given #424, we're not using cmake anymore, and instead we're opting for using the `cc` crate.

jdm:
@vyv03354 What was the motivation for this change? What C++ usage was prevented?

jdm:
Oh, this was literally just upstreaming the change from https://bugzilla.mozilla.org/show_bug.cgi?id=1451931?

PR #425, opened 2018-05-16
Get the mozjs crate to work with the mozjs_sys which uses bindgen

asajeffrey:
This is the matching PR to https://github.com/servo/mozjs/pull/134 which runs bindgen on spidermonkey to produce Rust bindings.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/425)


asajeffrey:
cc @jdm

asajeffrey:
This can land once https://github.com/servo/mozjs/pull/134 does.

asajeffrey:
https://crates.io/crates/mozjs_sys/0.51.0 is published, so we can land this.

@bors-servo r=jdm

PR #431, opened 2018-06-26
Update README.md to include yasm

jeremyjh:
ICU build fails without an assembler; `brew install yasm` resolves this.

```
DEBUG: configure: error: Building ICU requires either yasm or a GNU assembler. If you do not have either of those available for this platform you must use --without-intl-api
```

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/431)


KiChjang:
@bors-servo r+

Thanks!

PR #432, opened 2018-07-11
Add yasm as a dependency on MacOS

KiChjang:
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/432)


atouchet:
This is the same change as #431.

KiChjang:
Oh heh, closing.

PR #437, opened 2018-08-24
Update mozjs_sys to 0.61.0

SimonSapin:
Depends on https://github.com/servo/mozjs/pull/147

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/437)


SimonSapin:
Lost the race with publishing mozjs_sys on crates.io

@bors-servo retry

PR #445, opened 2018-11-06
Re-export IntoHandle

CYBAI:
When implementing `rejectionhandled` event, we'll need to add consumed rejection in task and we'll need to convert `HandleObject` into `RawHandleObject` which is `js::jsapi::HandleObject`. With exporting this trait, we can leverage the power of `into_handle` to do the conversion instead of writing a new method to get the `RawHandleObject`.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/445)

### DIFF
--- a/src/rust.rs
+++ b/src/rust.rs
@@ -68,7 +68,7 @@ use panic::maybe_resume_unwind;
 
 use default_heapsize;
 
-pub use mozjs_sys::jsgc::GCMethods;
+pub use mozjs_sys::jsgc::{GCMethods, IntoHandle};
 
 // From Gecko:
 // Our "default" stack is what we use in configurations where we don't have a compelling reason to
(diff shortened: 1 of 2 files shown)


src/rust.rs
```diff
@@ -68,7 +68,7 @@ use panic::maybe_resume_unwind;
 
 use default_heapsize;
 
-pub use mozjs_sys::jsgc::GCMethods;
+pub use mozjs_sys::jsgc::{GCMethods, IntoHandle};
```
CYBAI: We `use IntoHandle as IntoRawHandle` at line 12 (https://github.com/servo/rust-mozjs/pull/445/files#diff-34b82cd514c7429b32451dd0b07c1e1bR12).

Not sure I should `pub` there or re-export here?
But I think `pub` at line 12 will make the re-export trait name to be `IntoRawHandle` ?

PR #447, opened 2018-11-14
Introduce feature to only allow initialization once.

paulrouget:
See #22039

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/447)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,7 +2,7 @@
 name = "mozjs"
 description = "Rust bindings to the Mozilla SpiderMonkey JavaScript engine."
 repository = "https://github.com/servo/rust-mozjs"
-version = "0.9.3"
+version = "0.9.4"
 authors = ["The Servo Project Developers"]
 build = "build.rs"
 license = "MPL-2.0"
@@ -42,6 +42,7 @@ doctest = false
 
 [features]
 debugmozjs = ['mozjs_sys/debugmozjs']
+init_once = []
 
 [dependencies]
 lazy_static = "1"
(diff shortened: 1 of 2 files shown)


CYBAI:
servo/servo#22039

(Just help to reference the issue correctly)

Cargo.toml
```diff
@@ -2,7 +2,7 @@
 name = "mozjs"
 description = "Rust bindings to the Mozilla SpiderMonkey JavaScript engine."
 repository = "https://github.com/servo/rust-mozjs"
-version = "0.9.3"
+version = "0.9.4"
```
CYBAI: One thing I'm confused is that I've bumped the version to `0.9.4` at c16e7e2473 but GitHub didn't show it 🤔

https://github.com/servo/rust-mozjs/blob/c16e7e2473ff1203bc4510bed9d6618d14cc9d31/Cargo.toml#L5

paulrouget:
@CYBAI I can't see the commits from PR #445

CYBAI:
Hmm, @paulrouget can you see https://github.com/servo/rust-mozjs/pull/445/commits/9d5d8a9438e06fc025d927e12a1baf4b71647352 in the PR?

Btw, `mozjs` on `crates.io` is already on `0.9.4` (https://crates.io/crates/mozjs). Not sure why GitHub didn't show the commit 🤔

paulrouget:
Can't see the commit here: https://github.com/servo/rust-mozjs/commits/master/Cargo.toml but I see it here: https://github.com/servo/rust-mozjs/commits/master

Anyway. Gonna bump to 0.9.5 and publish.

nox:
Why not use `std::sync::Once`, and why is this feature only needed on Android in Servo? AFAIK SM should never be initialised multiple times, on all platforms, so I don't understand why this behaviour is enabled by a feature.

jdm:
It is specific to the way that we're using Servo inside Firefox Reality. The feature simply prevents us from calling shut down; it doesn't actually change how many times we initialize the library.

asajeffrey:
Hmm, might be worth finding another name for the feature? It seems to be more about shutdown than init, perhaps `never_shutdown` would be better?

jdm:
I have an open pull request that removes it entirely.

Issue #406, opened 2018-03-24
Add bindings for DataView typed array


jdm:
It will be very similar to the ArrayBufferView typed array bindings.

anthgur:
@jdm I think this might be a prerequisite to https://github.com/servo/servo/issues/20417 so I can take a crack at it too.

jdm:
So it looks like we can't reuse the macro because the JS API for DataView objects functions a little differently. We'll need to hand-write [this implementation](https://github.com/servo/rust-mozjs/blob/master/src/typedarray.rs#L268-L285) to use UnwrapArrayBuffer, [these functions](https://github.com/servo/mozjs/blob/3f38435122538e4b66f88878302f81331c0b6a44/mozjs/js/src/jsfriendapi.h#L2157-L2177), and [this implementation](https://github.com/servo/rust-mozjs/blob/master/src/typedarray.rs#L296-L307) using [this function](https://github.com/servo/mozjs/blob/3f38435122538e4b66f88878302f81331c0b6a44/mozjs/js/src/jsfriendapi.h#L2136-L2144). At that point we should be able to reuse the `array_alias!` macro

aditj:
@jdm  Can I work on this in a attempt to fix servo issue [#20417](https://github.com/servo/servo/issues/20417)

aditj:
> @jdm Can I work on this in an attempt to fix servo/servo issue [#20417](https://github.com/servo/servo/issues/20417)

Also if yes, can you maybe tell me if there a dxr tree for the same? :grin:  :sweat_smile:

jdm:
Absolutely. I'm pretty sure the answer to your DXR question is no, though, since the version of mozjs that we're using is at least 6 months older than the version in DXR/searchfox.

aditj:
@jdm Should I start by creating a macro rule for creating DataView element (like the one typed array element have)?

jdm:
Sure.

aditj:
@jdm 
I think the DataView element would be similar to the typed array element ,any properties that you think should be added?
Also, should I also write DataView Array structure? (And create a seperate array_alias macro for DataView Array)

jdm:
@aditj I do not believe we need a separate DataViewArray structure. We should be able to use the existing TypedArray structure without any changes; we just need to define an implementation of TypedArrayElement and TypedArrayElementCreator that uses the DataView APIs.

PR #453, opened 2019-02-21
Usable DataView typed array bindings 

aditj:
This PR tries to solve issue #406 
The changes use the traits of a typed array but the APIs for Dataview object from mozjs are used. 
The changes have not been tested , any suggested tests/changes would be appreciated. :)


jdm:
Can you add to the existing tests in https://github.com/servo/rust-mozjs/blob/master/tests/typedarray.rs? You can use the example at https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DataView as inspiration.

aditj:
We can't use the TypedArrayElement and TypedArrayElementCreator traits since the functions of length_and_data and create_new would be different in case of a DataView.

aditj:
> We can't use the TypedArrayElement and TypedArrayElementCreator traits since the functions of length_and_data and create_new would be different in case of a DataView.
@jdm Help  :face_with_head_bandage:  
Is there a way i could  overload functions of a trait (The same way I would overload functions in an OOP Language) , or is there any other alternative?

jdm:
Could you explain the problem that you are having? What is the code you are trying to write that is not working?

src/typedarray.rs
```diff
@@ -318,6 +321,53 @@ macro_rules! typed_array_element {
...
+
+    ($t: ident,
+     $element: ty,
+     $unwrap: ident,
+     $length: ident,
+     $create_new: ident,
+     $get_data: ident) => (
+        data_view_element!($t, $element, $unwrap, $length);
+        impl TypedArrayElementCreator for $t {
+            unsafe fn create_new(cx: *mut JSContext,array_buffer: *mut JSObject,byte_offset: u32,byte_length: u32) -> *mut JSObject {
```
aditj: @jdm 
This code  creates a create_new function for data view object by extending the TypedArrayElement , and due to the mismatch in definition and prototype we are having errors like (Link to travis CI errors):

1.  [for the length and data method](https://travis-ci.com/servo/rust-mozjs/jobs/179728105#L1152 ) 
2.  [for the create new method](https://travis-ci.com/servo/rust-mozjs/jobs/179728105#L1142)
etc.

aditj:
> Could you explain the problem that you are having? What is the code you are trying to write that is not working?

See this [comment](https://github.com/servo/rust-mozjs/pull/453/files#r261836062)

src/typedarray.rs
jdm: Ah, you're right. I don't know if we can reuse the typed array support code for data views after all :(

jdm:
I think we will have to create a new DataView type similar to [TypedArray](https://github.com/servo/rust-mozjs/blob/7811094fe96f4544e562844fc6bf10053bd10c85/src/typedarray.rs#L119-L122) which directly wraps [these APIs](https://github.com/servo/mozjs/blob/9b44b1e870cae2d96ea41bf2aa8a7d5eaa4cdce7/mozjs/js/src/jsfriendapi.h#L2229-L2279) and then add a macro like [typedarray!](https://github.com/servo/rust-mozjs/blob/7811094fe96f4544e562844fc6bf10053bd10c85/src/typedarray.rs#L413-L427) which creates a rooted instance.

aditj:
> rooted instance

I'll try doing that :+1:

PR #479, opened 2019-10-04
change the slot count to match the spidermonkey defintion

romankl:
right now the slot count is `2 + 38` in the spider monkey fork. this PR changes it to match the C++ definition.

The definition in C++: https://github.com/servo/mozjs/blob/dd175ee73a179ce218286c4be4ab80e1a75348be/mozjs/js/public/Class.h#L841

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/479)


jdm:
@bors-servo r+
Thanks!

PR #481, opened 2019-10-05
provide a basic readme with an example on how to eval a result in js

romankl:
this PR introduces new blocks in the readme to showcase how to eval
a string to a value.

BTW: the minor version is based on the latest published crates.io version. Should this be changed to the latest master? Or is a new release planned?

---
This change is [Reviewable](https://reviewable.io/reviews/servo/rust-mozjs/481)

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,4 +4,51 @@ Rust bindings to SpiderMonkey
 
 [Documentation](https://doc.servo.org/mozjs/)
 
-See https://github.com/servo/mozjs/blob/master/README.md for build instructions.
+
+# Usage
+
+Add this to your Cargo.toml:
+
+```toml
+[dependencies]
+mozjs = "0.10"
+```
+
+# Example
+
+```rust
+#[macro_use]
+extern crate mozjs;
+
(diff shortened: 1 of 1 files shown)


jdm:
We are no longer able to publish the mozjs crate (and transitively the rust-mozjs crate) to crates.io because of https://github.com/rust-lang/cargo/issues/6917. The documentation should not recommend using a published version because of this.

romankl:
@jdm do you think it makes sense to switch from the version-number-scheme to a Git-url? Or should this be removed?

jhwgh1968:
Hi @romankl, I wish I had seen your PR sooner. I just created #489, which does something similar in the crate doc string. The main difference is that mine refers to the online User Guide, and specifically translates two example blocks.

Your code helped me when I was tripping over some things in the Rust API. Could we merge our code somehow?

jdm:
My fear with adding this code to the readme in this form is that it will potentially get out of date when we upgrade spidermonkey, since it's not automatically compiled and validated in CI. I would rather add an `examples` directory that we can check with `cargo build --all-examples`.

PR #485, opened 2019-10-16
Throw when failing to convert sequence inner values

saschanaz:
Returning `Ok(ConversionResult::Failure)` allows the code to try other conversion in a union type, but [per the spec there is no such failover](https://heycam.github.io/webidl/#es-union) given that the type is iterable.

For https://github.com/servo/servo/issues/24406.


jdm:
@bors-setvo r+
Thanks!

CYBAI:
Oops, I just found that maybe we need a version update for publishing a new version to include this patch?

nox:
We don't really publish `mozjs_sys` and `mozjs` so shrug.

PR #487, opened 2019-11-23
Prevent initializing and shutting down SpiderMonkey from different threads

jdm:
Based on [this comment](https://github.com/servo/mozjs/blob/e21c05b415dfc246175ff8d5fc48b0e8c5b4e9e9/mozjs/js/public/Initialization.h#L19-L23), it appears that JS_Init and JS_ShutDown need to be called on the same thread. These changes make JSEngine a non-sendable type, and add JSEngineHandle as a sendable type that allows verifying that all outstanding consumers of the engine are dropped before shutting down.

### DIFF
--- a/src/rust.rs
+++ b/src/rust.rs
@@ -23,6 +23,7 @@ use std::os::raw::c_void;
 use std::cell::Cell;
 use std::marker::PhantomData;
 use std::sync::{Arc, Mutex};
+use std::sync::atomic::{AtomicU32, Ordering};
 
 use consts::{JSCLASS_RESERVED_SLOTS_MASK, JSCLASS_GLOBAL_SLOT_COUNT};
 use consts::{JSCLASS_IS_DOMJSCLASS, JSCLASS_IS_GLOBAL};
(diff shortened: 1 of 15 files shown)


jdm:
Note to self: need to update unit tests.

gterzian:
To elaborate on https://github.com/servo/servo/pull/24845#issuecomment-557863826

I think you could make `outstanding_handles: Arc<AtomicU32>` into a `outstanding_handles: Arc<Mutex<JsEngineHandles>, Condvar>`

with `JsEngineHandles` being:
```rust
enum JsEngineHandles {
    /// Initial state, equivalent to 0 handles, but before any handles have been created.
    Initialized,
    /// Count handles, start at 1, goes to 0 when all handles are dropped.
    Handles(u32)
}
```

with `can_shutdown` containing the equivalent of  `while self.outstanding_handles != JsEngineHandles::Handes(0) {//wait on the condvar}`

src/rust.rs
```diff
@@ -172,11 +173,31 @@ pub enum JSEngineError {
 /// A handle that must be kept alive in order to create new Runtimes.
 /// When this handle is dropped, the engine is shut down and cannot
 /// be reinitialized.
-pub struct JSEngine(());
+pub struct JSEngine {
+    /// The count of alive handles derived from this initialized instance.
+    outstanding_handles: Arc<AtomicU32>,
+    // Ensure this type cannot be sent between threads.
```
gterzian: Could `impl !Send for JSEngine{}` be used for that purpose?

jdm:
```
error[E0658]: negative trait bounds are not yet fully implemented; use marker types for now
   --> src/rust.rs:181:1
    |
181 | impl !Send for JSEngine {}
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: for more information, see https://github.com/rust-lang/rust/issues/13231
    = help: add `#![feature(optin_builtin_traits)]` to the crate attributes to enable
```
I'd rather not introduce nightly-only features to this crate.

jdm:
The idea of making shutdown be a blocking operation is interesting, but I think I'd rather not make that a requirement of using this crate. It's not difficult to imagine a consumer writing code that uses this API that ends up creating a deadlock by not ensuring that all of the engine handles are disposed of before attempting to shutdown. I would rather provide an API that panics by default and allow consumers to build their own blocking shutdown behaviour on top of it, like https://github.com/servo/servo/pull/24862 does for Servo.

gterzian:
>I would rather provide an API that panics by default and allow consumers to build their own blocking shutdown behaviour on top of it

That makes sense.

PR #489, opened 2019-12-01
Add crate docstring and two code examples

jhwgh1968:
In the hopes of embedding SpiderMonkey into a non-Servo Rust project, I spent some time digging into these bindings. Unfortunately, it was very easy to get lost.

This PR is my attempt to make it easier for the next person who attempts my exercise by adding a few sign posts I would have found helpful, and translating one of the basic examples from the SpiderMonkey User Guide on MDN to Rust. Documenting the entire crate (and where it dips directly into ``mozjs_sys``) is a task too big for any one person, I think.

EDIT: removed WIP comment

### DIFF
--- a/examples/eval.rs
+++ b/examples/eval.rs
@@ -1,0 +1,62 @@
+#![allow(
+    non_upper_case_globals,
+    non_camel_case_types,
+    non_snake_case,
+    improper_ctypes
+)]
+
+//! # Running scripts
+//! Here is the code under "Running scripts" in the MDN User Guide[1] translated into Rust. This
+//! only shows the ``run()`` function's contents because the original does as well.
+//!
+//! The actual code that is run is designed to be testable, unlike the example given.
+//! [1]: https://developer.mozilla.org/en-US/docs/Mozilla/Projects/SpiderMonkey/JSAPI_User_Guide
+//!
+
+use std::ptr;
+
+#[macro_use]
+extern crate mozjs;
+use mozjs::jsapi::*;
+use mozjs::jsval::UndefinedValue;
+use mozjs::rust::SIMPLE_GLOBAL_CLASS;
+use mozjs::rust::{JSEngine, RealmOptions, Runtime};
+
+fn run(rt: Runtime) {
+    let options = RealmOptions::default();
+    rooted!(in(rt.cx()) let global = unsafe {
+        JS_NewGlobalObject(rt.cx(), &SIMPLE_GLOBAL_CLASS, ptr::null_mut(),
+                           OnNewGlobalHookOption::FireOnNewGlobalHook,
+                           &*options)
+    });
+
+    /* These should indicate source location for diagnostics. */
+    let filename: &'static str = "inline.js";
+    let lineno: u32 = 1;
+
+    /*
+     * The return value comes back here. If it could be a GC thing, you must add it to the
+     * GC's "root set" with the rooted! macro.
+     */
+    rooted!(in(rt.cx()) let mut rval = UndefinedValue());
+
+    /*
+     * Some example source in a string. This is equivalent to JS_EvaluateScript in C++.
+     */
+    let source: &'static str = "40 + 2";
+
+    let res = rt.evaluate_script(global.handle(), source, filename, lineno, rval.handle_mut());
+
+    if res.is_ok() {
+        /* Should get a number back from the example source. */
+        assert!(rval.get().is_int32());
+        assert_eq!(rval.get().to_int32(), 42);
+    }
+}
+
+fn main() {
+    let engine = JSEngine::init().expect("failed to initalize JS engine");
+    let runtime = Runtime::new(engine.handle());
+    assert!(!runtime.cx().is_null(), "failed to create JSContext");
+    run(runtime);
+}
--- a/examples/minimal.rs
+++ b/examples/minimal.rs
@@ -1,0 +1,52 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#![allow(
+    non_upper_case_globals,
+    non_camel_case_types,
+    non_snake_case,
+    improper_ctypes
+)]
+
+//! # A minimal example
+//! Here is the code under "A minimal example" in the MDN User Guide[1] translated into Rust.
+//! [1]: https://developer.mozilla.org/en-US/docs/Mozilla/Projects/SpiderMonkey/JSAPI_User_Guide
+
+use std::ptr;
+
+#[macro_use]
+extern crate mozjs;
+use mozjs::rust::SIMPLE_GLOBAL_CLASS;
+use mozjs::{jsapi::*, rust::JSEngine, rust::RealmOptions, rust::Runtime};
+
+fn main() {
+    // Initialize the JS engine. This handle must be kept alive in order to create new Runtimes.
+    let engine = JSEngine::init().expect("failed to initalize JS engine");
+
+    // Create a Runtime -- wraps a JSContext in the C++ API.
+    let runtime = Runtime::new(engine.handle());
+    assert!(!runtime.cx().is_null(), "failed to create JSContext");
+
+    run(runtime);
+
+    // There is no need for the shut down block in the C++, because rust destructors and Arc
+    // reference counts will clean up everything.
+}
+
+fn run(rt: Runtime) {
+    let cx = rt.cx();
+    // In addition to what the C++ interface requires, define a global scope for the code.
+    //
+    // This demonstrates the way Rust uses the C++ garbage collector: using the rooted! macro to
+    // indicate when the GC can collect them.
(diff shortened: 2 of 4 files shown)


jhwgh1968:
I've fought with this quite a bit, but it's really a struggle. I can't figure out how to make a global object (and a null pointer anywhere results in a crash), and the API is really hard to use. My attempts to figure out how Servo uses this library -- the only working code I know of -- seem futile.

Surely I am looking at this API upside down and this is not as difficult as it seems. Any guidance on offer? What am I missing?

jhwgh1968:
It's still a bit kludgy, I think, but should work and demonstrate the basics of the API pretty well. Ready for review.

src/lib.rs
```diff
@@ -7,6 +7,125 @@
...
+//! This crate contains Rust bindings to the [SpiderMonkey Javascript engine][1]
+//! developed by Mozilla.
+//!
+//! These bindings are designed to be a fairly straightforward translation to the C++ API, while
+//! taking advantage of Rust's memory safety. For more about the Spidermonkey API, see the
+//! [API Reference][2] and the [User Guide][3] on MDN.
+//!
+//! [1]: https://developer.mozilla.org/en-US/docs/Mozilla/Projects/SpiderMonkey
+//! [2]: https://developer.mozilla.org/en-US/docs/Mozilla/Projects/SpiderMonkey/JSAPI_reference
+//! [3]: https://developer.mozilla.org/en-US/docs/Mozilla/Projects/SpiderMonkey/JSAPI_User_Guide
```
jdm: Let's add a link to https://github.com/mozilla-spidermonkey/spidermonkey-embedding-examples/ as well.

src/lib.rs
```diff
@@ -7,6 +7,125 @@
...
+//!
+//! # A minimal example
+//! Here is the code under "A minimal example" in the User Guide translated into Rust:
+//!
+//! ```rust
+//! use std::ffi::CStr;
```
jdm: Per my comment in https://github.com/servo/rust-mozjs/pull/481#issuecomment-577588059, can we turn this into an actual example file rather than a docstring?

src/lib.rs
```diff
@@ -7,6 +7,125 @@
...
+//! ```
+//!
+//! # Running scripts
+//! Here is the translated code for "Running scripts" lower on the same page. This only shows the
+//! ``run()`` function's contents because the original does as well.
+//!
+//! The actual code that is run is designed to be testable, unlike the example given.
+//!
+//! ```rust
+//! # use std::ffi::CStr;
```
jdm: This would benefit from being a separate example file as well.

jhwgh1968:
Thanks, @jdm. I didn't know CI didn't run doc tests. Given that, I agree that examples make sense.

I could not figure out how to make the examples get their own doc pages, so did my best to generate links to the source files in the crate root, since I feel those should be mentioned. If you know of a trick or prefer I remove them, let me know.

jhwgh1968:
(Note to self: there is a request review button at the bottom, when GitHub doesn't detect that I have fixed all the changes sometimes.)

examples/eval.rs
```diff
@@ -0,0 +1,59 @@
...
+
+        /* These should indicate source location for diagnostics. */
+        let filename: &'static str = "inline.js";
+        let lineno: u32 = 1;
+
+        /*
+         * The return value comes back here -- if it could be a GC thing, you must
+         * add it to the GC's "root set" with the rooted! macro.
+         */
+        let mut rval: Value = std::mem::zeroed();
```
jdm: We should be using the rooted! macro here to avoid unsafety.
```rust
rooted!(in(rt.cx()) let mut rval = UndefinedValue());
```
Then we'll use `rval.handle_mut()` later, and `rval.get()`.

examples/minimal.rs
```diff
@@ -0,0 +1,48 @@
...
+    // There is no need for the shut down block in the C++, because rust destructors and Arc
+    // reference counts will clean up everything.
+}
+
+fn run(rt: Runtime) {
+    let cx = rt.cx();
+    // In addition to what the C++ interface requires, define a global scope for the code.
+    //
+    // This demonstrates the way Rust uses the C++ garbage collector: using the rooted! macro to
+    // indicate when the GC can collect them.
```
jdm: Since we don't use the macro at all in this example, I don't think this sentence makes sense.

jhwgh1968: Uh, I use the ``rooted!`` macro on line 41... but I think I am missing something bigger. See comment below.

jhwgh1968:
Thanks, @jdm! I have fixed one of the comments, but the other one tells me that I am missing something. I have two questions.

The line you commented on ``minimal.rs`` was a bit of a mess. However, that mess was trying to copy from the C++ example, which was specifically showing *how to expose a global to SpiderMonkey*. I used the ``rooted!`` macro combined with ``JS_NewGlobalObject`` basically out of trial and error. I could not find anything else that would work in RustDoc.

I presume creating a global like this in Rust is *intentionally* difficult, because Rust wants in-depth knowledge of lifetimes. Thus, my first question: does this mean that is a bad example for Rust? Or is there a better way to do this I missed?

Regarding ``eval.rs``, I could not remove the ``unsafe`` block even after implementing your suggestion. That is because the same ``JS_NewGlobalObject`` construction is in there, too, and the ``global`` variable would not live long enough otherwise.

Why is it is there? Because I read a "global root object" is used as a compartment -- at least, in the C++ version. In Rust, I only know that if I mess up that object creation, lots of ugly UB happens.

That is my second question: does that need to be in ``eval.rs`` at all? A lot of the other code like ``AutoRequest`` is dropped in the translation to Rust, so can that go out as well and I just missed it?

Thanks for your patience!

examples/minimal.rs
jdm: Sorry, I don't know how I missed that :<

examples/eval.rs
```diff
@@ -0,0 +1,60 @@
...
+extern crate mozjs;
+use mozjs::jsapi::*;
+use mozjs::jsval::UndefinedValue;
+use mozjs::rust::{JSEngine, RealmOptions, Runtime};
+use mozjs::rust::SIMPLE_GLOBAL_CLASS;
+
+fn run(rt: Runtime) {
+    unsafe {
+        let options = RealmOptions::default();
+        rooted!(in(rt.cx()) let global =
```
jdm: We can probably do
```rust
rooted!(in(rt.cx()) let global = unsafe {
    JS_NewGlobalObject(...)
});
```
so the scope of the unsafety is reduced.

examples/minimal.rs
```diff
@@ -0,0 +1,48 @@
...
+    unsafe {
```
jdm: Same here.

examples/eval.rs
jhwgh1968: I thought I tried that, but apparently not. Thanks!

jhwgh1968:
Thanks for the pointers and comments, @jdm! All fixed!

I also ran Rustfmt on it to make sure it's nice and clean. :sparkles:

jhwgh1968:
Uh... the AppVeyor run seems to have a broken mozjs_sys on win32:
```
: failed to run custom build command for `mozjs_sys v0.67.1 (https://github.com/servo/mozjs?rev=bc4935b668171863e537d3fb0d911001a6742013#bc4935b6)`
Caused by:
  process didn't exit successfully: `C:\projects\rust-mozjs\target\debug\build\mozjs_sys-0c301899e9ba10f4\build-script-build` (exit code: 101)
--- stdout
cargo:outdir=C:\projects\rust-mozjs\target\i686-pc-windows-msvc\debug\build\mozjs_sys-66162aad09d2f7b9\out\build
[[ 'C:\Users\appveyor\.cargo\git\checkouts\mozjs-fa11ffc7d4f1cc2d\bc4935b\mozjs'/js/src/configure -ot 'C:\Users\appveyor\.cargo\git\checkouts\mozjs-fa11ffc7d4f1cc2d\bc4935b\mozjs'/js/src/configure.in ]] && touch 'C:\Users\appveyor\.cargo\git\checkouts\mozjs-fa11ffc7d4f1cc2d\bc4935b\mozjs'/js/src/configure || true
C:\Users\appveyor\.cargo\git\checkouts\mozjs-fa11ffc7d4f1cc2d\bc4935b\makefile.cargo:191: recipe for target 'maybe-configure' failed
--- stderr
      0 [main] us 0 init_cheap: VirtualAlloc pointer is null, Win32 error 487
AllocationBase 0x0, BaseAddress 0x607A0000, RegionSize 0x320000, State 0x10000
C:\mozilla-build\msys\bin\sh.exe: *** Couldn't reserve space for cygwin's heap, Win32 error 0
mozmake: *** [maybe-configure] Error 1
thread 'main' panicked at 'assertion failed: result.success()', C:\Users\appveyor\.cargo\git\checkouts\mozjs-fa11ffc7d4f1cc2d\bc4935b\build.rs:167:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```

I can't imagine it's related to my simple change...

jdm:
Yeah, that's a known intermittent error that occurs.

PR #494, opened 2020-01-22
Update SpiderMonkey

nox:


nox:
The `enumerate` test seems to fail on x86_64 on Windows.

nox:
Even better, this does not build on Windows with `--features debugmozjs`.

nox:
Error seems to be https://github.com/rust-lang/libc/pull/1433

appveyor.yml
```diff
@@ -30,16 +31,18 @@ environment:
...
-  LINKER: "lld-link.exe"
 
   VCVARSALL_PATH: C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Auxiliary\Build
 
   matrix:
     - TARGET: i686-pc-windows-msvc
       CROSS_COMPILE: 1
     - TARGET: x86_64-pc-windows-msvc
+    - TARGET: x86_64-pc-windows-msvc
+      FEATURES: --features debugmozjs
```
jdm: The reason we didn't have this job before is because of https://github.com/servo/mozjs/issues/189.

nox: Yeah I figured that out but I wanted to see if a debug build would tell me more about the `enumerate` test failing. Maybe I fumbled updating the C++ glue and I badly copied a prototype somewhere.

jdm:
I am 99% sure this problem is happening because our JSValue creation and extraction code no longer matches the C++ methods. Compare https://searchfox.org/mozilla-central/source/js/public/Value.h#798-812 against:
* https://github.com/servo/mozjs/blob/master/src/jsval.rs#L182-L196
* https://github.com/servo/mozjs/blob/bc4935b668171863e537d3fb0d911001a6742013/src/jsval.rs#L436-L449
* https://github.com/servo/rust-mozjs/blob/master/src/jsval.rs#L181-L195

The invalid pointer that I observe when the segfault occurs matches the valid pointer shifted right one bit. The reason this only affects the JIT is that all of our Rust code using `.toPrivate()` performs the shifting that we did when creating values with `PrivateValue(..)`, but the JIT relies on the C++ implementations which do not.

nox:
Will write a fix first thing in the morning if you didn't already by then. Thank you for your help, really.

nox:
```
[mozjs 0.13.0] running: "clang-cl.exe" "-nologo" "-MD" "-Z7" "-Brepro" "-m64" "-I" "C:\\projects\\rust-mozjs\\target\\x86_64-pc-windows-msvc\\debug\\build\\mozjs_sys-348968a14b8060be\\out\\build\\dist\\include" "-W4" "-MDd" "-Od" "-FIC:\\projects\\rust-mozjs\\target\\x86_64-pc-windows-msvc\\debug\\build\\mozjs_sys-348968a14b8060be\\out\\build\\js\\src\\js-confdefs.h" "-Zi" "-GR-" "-Wno-c++0x-extensions" "-Wno-return-type-c-linkage" "-Wno-invalid-offsetof" "-Wno-unused-parameter" "-DDEBUG=" "-D_DEBUG=" "-DWIN32=" "-FoC:\\projects\\rust-mozjs\\target\\x86_64-pc-windows-msvc\\debug\\build\\mozjs-83b0ce056021e4ff\\out\\src/jsglue.o" "-c" "src/jsglue.cpp"
```

Now to find where that `-MDd` is from…

jdm:
Keep in mind that https://github.com/servo/mozjs/commit/2b97c2dc648c84164a6a12418dce0e9cd3b62038 exists, if that helps or hinders at all.

nox:
Thanks for the pointer. The solution AFAIK was to set `MOZ_NO_DEBUG_RTL` in `mozjs_sys` and remove `-MDd` here.

nox:
I suspect some sort of ABI issue on Windows with `GetPropertyKeys`.

nox:
I fixed the failing `enumerate` test, this was indeed an ABI issue with MSVC (and clang-cl it seems).

nox:
I fixed the failing `enumerate` test, this was indeed an ABI issue with MSVC (and clang-cl it seems).

nox:
AFAICT, we do not include `js-config.h` when building MFBT, we use `js-confdefs.h`, which seems to only include configure variables from the `old-configure` script, not the fancy Python ones we define ourselves.

appveyor.yml
```diff
@@ -30,16 +31,18 @@ environment:
...
   CXX: "clang-cl.exe"
-  LINKER: "lld-link.exe"
 
   VCVARSALL_PATH: C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Auxiliary\Build
 
   matrix:
     - TARGET: i686-pc-windows-msvc
       CROSS_COMPILE: 1
     - TARGET: x86_64-pc-windows-msvc
+    - TARGET: x86_64-pc-windows-msvc
```
asajeffrey: Should we add some UWP targets too?

src/consts.rs
```diff
@@ -2,26 +2,17 @@
...
-    1 << (JSCLASS_HIGH_FLAGS_SHIFT + 1);
+pub use jsapi::JSCLASS_IS_GLOBAL;
 
-pub const JSCLASS_IS_PROXY: c_uint =
-    1 << (JSCLASS_HIGH_FLAGS_SHIFT + 4);
+pub use jsapi::JSCLASS_IS_PROXY;
 
-pub const JSCLASS_GLOBAL_SLOT_COUNT: c_uint =
-    JSCLASS_GLOBAL_APPLICATION_SLOTS + JSProtoKey::JSProto_LIMIT as c_uint * 2 + 38;
+pub use jsapi::JSCLASS_GLOBAL_SLOT_COUNT;
```
asajeffrey: Yay for bindgenup!

src/rust.rs
```diff
@@ -379,19 +376,20 @@ impl Runtime {
...
 
         unsafe {
-            if !EvaluateUtf8(self.cx(), options.ptr, ptr as *const _, len as size_t, rval.into()) {
+            let mut source = SourceText {
+                units_: script.as_ptr() as *const _,
+                length_: script.len() as u32,
+                ownsUnits_: false,
+                _phantom_0: std::marker::PhantomData,
+            };
+            if !Evaluate2(self.cx(), options.ptr, &mut source, rval.into()) {
```
asajeffrey: We don't need the `Utf` marker any more?

src/rust.rs
```diff
@@ -379,19 +376,20 @@ impl Runtime {
...
         let filename_cstr = ffi::CString::new(filename.as_bytes()).unwrap();
         debug!("Evaluating script from {} with content {}", filename, script);
-        // SpiderMonkey does not approve of null pointers.
-        let (ptr, len) = if script.len() == 0 {
-            static empty: &'static [u8] = &[];
-            (empty.as_ptr(), 0)
-        } else {
-            (script.as_bytes().as_ptr(), script.len() as c_uint)
-        };
-        assert!(!ptr.is_null());
```
asajeffrey: It got over its dislike of `null`?

appveyor.yml
jdm: Can't without using something like xargo to also build the standard library, sadly.

src/rust.rs
nox: `source` is a  `root::JS::SourceText<root::mozilla::Utf8Unit>`, `Evaluate2` only understands UTF-8.

src/rust.rs
nox: Empty slices have not been the `null` pointer for years.

nox: And strings etc I mean, an `&str` with a `null` pointer is UB and has been for many years.

PR #497, opened 2020-03-08
Expose a function to transform slicer into SourceText

CYBAI:
Add this function so that we can reuse it in mozjs and servo.

### DIFF
--- a/src/rust.rs
+++ b/src/rust.rs
@@ -57,6 +57,7 @@ use jsapi::MutableHandleIdVector as RawMutableHandleIdVector;
 use jsapi::glue::{JS_Init, JS_NewRealmOptions, DeleteRealmOptions};
 #[cfg(feature = "debugmozjs")]
 use jsapi::mozilla::detail::GuardObjectNotificationReceiver;
+use jsapi::mozilla::Utf8Unit;
 
 use jsval::ObjectValue;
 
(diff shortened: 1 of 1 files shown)


src/rust.rs
```diff
@@ -249,6 +249,15 @@ impl Drop for JSEngine {
     }
 }
 
+pub fn to_source_text<T, U>(source: &[T]) -> SourceText<U> {
+    SourceText {
+        units_: source.as_ptr() as *const _,
+        length_: source.len() as u32,
```
CYBAI: @nox after trying to use the `&str` as argument version in Servo, I found it's not so convenient to pass a string slice because we might want to `encode_utf16` first.

Then, I tried to change it to `fn to_source_text<T>(source: &[T]) -> SourceText<T>`; however, while passing `&[u8]`, we can't return a `SourceText<u8>`. I think that's what you were talking about we can't just pass the wrapper into `Evaluate2`.

So, I changed the type to current version 🤔 but I wonder, is it possible to have some type level restriction to avoid some misusing here?

Or, I'm wondering is it possible or better to define `impl From` for `SourceText` 🤔?

nox: I would just make two functions, returning 2 different `SourceText<_>`.

CYBAI: oh! it sounds good to me so that we can restrict the type for `u8` and `u16`, thanks!

nox: The entry for the UTF-8 one should be a `&str`, not an `&[u8]`. And the output should be a `SourceText<mozilla::Utf8Unit>`.

CYBAI: Updated to fix! Should be better now 👀 Thanks!

CYBAI:
ping @nox 👀 ?

PR #499, opened 2020-03-27
Update mozjs.

jdm:


sagudev:
The cause of error is https://bugzilla.mozilla.org/show_bug.cgi?id=1616312

PR #508, opened 2020-04-28
Expose binding for JS::DescribeScriptedCaller.

### LINKED ISSUE #507: Implement a wrapper around DescribeScriptedCaller
https://searchfox.org/mozilla-central/rev/158bac3df3a1890da55bdb6ffdaf9a7ffc0bfb0a/js/src/jsapi.h#2783-2804

This API takes a JS::AutoFilename argument, and implementing a Rust wrapper around just that type would require duplicating C++ logic in https://searchfox.org/mozilla-central/rev/158bac3df3a1890da55bdb6ffdaf9a7ffc0bfb0a/js/src/jsapi.cpp#5512-5549 which is a waste. Much better to write a small C++ stub that copies the filename string out into a buffer that Rust provides instead.

jdm:
Fixes #507.

### DIFF
--- a/src/rust.rs
+++ b/src/rust.rs
@@ -18,6 +18,7 @@ use std::slice;
 use std::str;
 use std::u32;
 use std::default::Default;
+use std::ffi::CStr;
 use std::ops::{Deref, DerefMut};
 use std::os::raw::c_void;
 use std::cell::Cell;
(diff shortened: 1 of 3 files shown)


src/rust.rs
```diff
@@ -1297,6 +1298,28 @@ macro_rules! new_jsjitinfo_bitfield_1 {
...
 
+#[derive(Debug, Default)]
+pub struct ScriptedCaller {
+    pub filename: String,
+    pub line: u32,
+    pub col: u32,
+}
+
+pub unsafe fn describe_scripted_caller(cx: *mut JSContext) -> Result<ScriptedCaller, ()> {
+    let mut buf = [0; 1024];
```
Manishearth, line 1309: seems a bit big but i don't think it matters

PR #511, opened 2020-05-25
Incrementing minor version for underlying source API

gterzian:
I forgot to do this in https://github.com/servo/rust-mozjs/pull/495, I guess it is necessary in order to update in Servo?


gterzian:
@jdm r?

jdm:
We use a git dependency, so this isn't necessary.

PR #512, opened 2020-05-25
Use dynamic dispatch in `mozjs::panic::wrap_panic`

SimonSapin:
Servo has generated code that calls it with thousands of different closures. With monomorphization, this function alone contributed significantly to code size (millions of lines of unoptimized LLVM IR) and compilation time: https://github.com/servo/servo/issues/26585

By switching to dynamic dispatch, compilation time for Servo’s script crate (which is pathologically long) is reduced by about a quarter on my machine.

Other changes are:

* Removing return value handling, which required a generic type parameter. Instead, the provided callback needs to assign to a captured variable. (See test for an example.)

* The callback needs to implement `FnMut` rather than only `FnOnce`. We could instead take `Box<dyn FnOnce() + '_>` (with allocation overhead), but none of the callers is affected.

* The `UnwindSafe` constraint is removed, because multi-trait objects do not support up-casting yet. It wasn’t much help anyway  since the code generator always used `AssertUnwindSafe` regardless of what each closure does.

Dynamic dispatch has non-zero runtime cost. Perhaps here it is small enough, or compensated by better instruction cache utilization thanks to reduced code size?

### DIFF
--- a/src/panic.rs
+++ b/src/panic.rs
@@ -4,7 +4,7 @@
 
 use std::any::Any;
 use std::cell::RefCell;
-use std::panic::{UnwindSafe, catch_unwind, resume_unwind};
+use std::panic::{AssertUnwindSafe, catch_unwind, resume_unwind};
 
 thread_local!(static PANIC_PAYLOAD: RefCell<Option<Box<dyn Any + Send>>> = RefCell::new(None));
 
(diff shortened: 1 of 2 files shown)


jdm:
@bors-servo r-
Sorry, jumped the gun.

jdm:
I'm restarting the single job that timed out.

PR #515, opened 2020-07-18
Add glue methods to get JSValue of a private reference

CYBAI:
While working on servo/servo#27026, we've noticed calling
`GetScriptPrivate` will hit a CRASH consistently in Windows. jdm noticed
that this is caused by "can't return JSValue by value on Windows"; thus,
we need to expose these 2 glue methods so that we can handle the
`GetScriptPrivate` and `GetModulePrivate` correctly.

---

This will help servo/servo#27026.

We might want to handle the `ModulePrivate` much more correctly in the future so it might be worth adding it here as well?

### DIFF
--- a/src/jsglue.cpp
+++ b/src/jsglue.cpp
@@ -19,6 +19,7 @@
 #include "js/Class.h"
 #include "js/Id.h"
 #include "js/MemoryMetrics.h"
+#include "js/Modules.h"  // include for JS::GetModulePrivate
 #include "js/Principals.h"
 #include "js/Promise.h"
 #include "js/Proxy.h"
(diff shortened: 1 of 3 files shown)


CYBAI:
Btw, I just noticed the Microsoft documentation link in [line 1072](https://github.com/servo/rust-mozjs/pull/515/files#diff-0c2f4ced2d2186c7a57c783c523a42c1R1072) is 404 now.

I tried to search it but looks like I can't find exactly same one.

- https://docs.microsoft.com/en-us/cpp/code-quality/annotating-function-parameters-and-return-values?view=vs-2019
- https://docs.microsoft.com/en-us/cpp/cpp/reference-type-function-returns?view=vs-2019

But found 2 links might be related; ☝️ is any of them I should update to the comment?

jdm:
Please file an issue about the broken doc link.

src/jsglue.cpp
```diff
@@ -1077,6 +1078,16 @@ JS_GetPromiseResult(JS::HandleObject promise, JS::MutableHandleValue dest) {
   dest.set(JS::GetPromiseResult(promise));
 }
 
+void
+JS_GetScriptPrivate(JSScript* script, JS::Value* dest) {
```
jdm, line 1082: We should use `JS::MutableHandleValue dest` like JS_GetPromiseResult (and that will match what's declared in glue.rs as well).

src/jsglue.cpp
```diff
@@ -1077,6 +1078,16 @@ JS_GetPromiseResult(JS::HandleObject promise, JS::MutableHandleValue dest) {
...
+    *dest = JS::GetScriptPrivate(script);
+}
+
+void
+JS_GetModulePrivate(JSObject* module, JS::Value* dest) {
```
jdm, line 1087: Same here.

src/jsglue.cpp
CYBAI: > and that will match what's declared in glue.rs as well

Indeed! Thanks for catching this! Updated!

PR #520, opened 2020-07-27
CompileOptionsWrapper::new should accept a &str instead of a C string pointer

### LINKED ISSUE #518: CompileOptionsWrapper::new should accept a &str instead of a C string pointer
It's too easy to misuse, otherwise. We can do the rust->c string conversion inside the constructor.

nicoabie:
Updated CompileOptionsWrapper new method to receive a &str for the filename

Also changed the line parameter to u32 instead of c_int (to make the bits clearer)

Fixes #518

### DIFF
--- a/src/rust.rs
+++ b/src/rust.rs
@@ -396,12 +396,11 @@ impl Runtime {
     pub fn evaluate_script(&self, glob: HandleObject, script: &str, filename: &str,
                            line_num: u32, rval: MutableHandleValue)
                     -> Result<(),()> {
-        let filename_cstr = ffi::CString::new(filename.as_bytes()).unwrap();
         debug!("Evaluating script from {} with content {}", filename, script);
 
         let _ac = JSAutoRealm::new(self.cx(), glob.get());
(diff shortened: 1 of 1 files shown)


src/rust.rs
```diff
@@ -396,7 +396,6 @@ impl Runtime {
     pub fn evaluate_script(&self, glob: HandleObject, script: &str, filename: &str,
                            line_num: u32, rval: MutableHandleValue)
                     -> Result<(),()> {
-        let filename_cstr = ffi::CString::new(filename.as_bytes()).unwrap();
```
CYBAI, line 399: Because this variable is removed, the `filename_cstr` is still used in line 403 so CI failed.

I think one more thing we need to do here is to use `filename` directly in line 403.

https://github.com/servo/rust-mozjs/blob/d88435c3664edb280eca2f6b2610285892545839/src/rust.rs#L402-L404

nicoabie: Thanks a lot for the feedback @CYBAI

PR #523, opened 2020-08-25
rust.rs: unsafe Handle::new() is now a private function.

sudarshan-reddy:
The new() constructor is unsafe because of the way it accepts pointers 
and isn't marked to indicate this. The usage occurrence of this function 
outside of this crate is also limited to one. To thisend, marking it private 
would prevent potentially unsafe propagation.


jdm:
@bors-servo r+
Thanks!

PR #536, opened 2021-03-12
Update mozjs_sys revision

ksqsf:
The old commit ID seems to be blocking Servo from building on Xcode 11.1+. Propose to update the revision ID to be [more recent](https://github.com/servo/mozjs/commit/1e1c40f4d74445208074ef322a6722fda9757433).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -58,4 +58,4 @@ lazy_static = "1"
 libc = "0.2"
 log = "0.4"
 num-traits = "0.2"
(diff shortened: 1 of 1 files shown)


ksqsf:
Seems already covered by #535 !

ksqsf:
So #535 [is not ready yet](https://github.com/servo/rust-mozjs/pull/535/files#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542R61), reopen.

PR #552, opened 2021-09-05
Added Test for Property Descriptor

Redfire75369:
This pull request adds a test for functions using property descriptors.

This test was created on request of @sagudev. I believe the request was to allow for testing when changes in the JSAPI that occurred in SM 90/91. The changes caused these functions to use `mozilla::Maybe`.


jdm:
Sorry for not noticing this!
@bors-servo r+

PR #556, opened 2022-02-20
Update SpiderMonkey to 97

CYBAI:
The mozjs PR is at https://github.com/servo/mozjs/pull/295

---

- [ ] Change the mozjs_sys back to the servo one with new revision in `Cargo.toml`


src/jsglue.cpp
```diff
@@ -1062,13 +1087,13 @@ CallUnbarrieredObjectTracer(JSTracer* trc, JSObject** objp, const char* name)
 void
 CallObjectRootTracer(JSTracer* trc, JSObject** objp, const char* name)
 {
-    JS::UnsafeTraceRoot(trc, objp, name);
+    JS::TraceRoot(trc, objp, name);
```
CYBAI, line 1090: This method was renamed at https://bugzilla.mozilla.org/show_bug.cgi?id=1744059.

